PR #3, opened 2015-07-24
Use copy artifacts instead of clone workspace;

Ooblioob:
the clone workspace plugin was extremely inefficient and took
several minutes to clone a workspace.  Rewrote the same functionality
using the copy artifacts plugin to give us a significant
performance improvement.


OrlandoSoto:
:+1:

imuchnik:
:+1:

PR #8, opened 2015-09-25
JS Job Builder

imuchnik:
First stab at using builder vs templates.


Ooblioob:
Love it.  Pattern is simple enough we can still teach it to other developers but solves the particularly onerous problem of passing variables to templates.   

One downside here is that every seed job will need to include the Jenkins-automation repo, but I don't see this as a big problem.  

:+1: ready to merge

PR #9, opened 2015-10-01
BDD Security Template; Add log rotation

Ooblioob:
Created a template for using BDD Security.  This is based on our implementation of it, and does require us to use an environment variable to an internal repository.  I've added a comment with a link to the open sourced version of BDD security for anyone who wants to use that instead, but it was the only solution I could find to the problem.  Suggestions welcome.  

Also set the log rotation to something reasonable. I think this is probably one area where we simply forget, so having it in the base template is at least helpful.  

@imuchnik Can you please review?

### DIFF
--- a/jenkins-automation.groovy
+++ b/jenkins-automation.groovy
@@ -197,15 +197,15 @@ job('template-bdd-security'){
   }
 
   steps {
-    shell("""
+    shell("""\
 umask 002
 
 /usr/bin/Xvfb :1 -ac -screen 0 1024x768x24 &
 sleep 10
 export DISPLAY=:1
 
-sed -i 's/<zapPath>.*<\/zapPath>/<zapPath>\/var\/lib\/jenkins\/workspace\/'\${JOB_NAME}'\/zap\/zap.sh<\/zapPath>/g' config.xml
-sed -i 's/<baseUrl><\/baseUrl>/<baseUrl>\$BASE_URL<\/baseUrl>/g' config.xml
+sed -i 's/<zapPath>.*<\\/zapPath>/<zapPath>\\/var\\/lib\\/jenkins\\/workspace\\/'\${JOB_NAME}'\\/zap\\/zap.sh<\\/zapPath>/g' config.xml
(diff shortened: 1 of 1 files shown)


jenkins-automation.groovy
```diff
@@ -203,35 +264,36 @@ buildFlowJob('template-base-build-flow') {
...
+//   steps {
+//     dsl {
+//       external(['**/jenkins.groovy'])
+//       additionalClasspath('lib/*.jar')
+//       removeAction('DISABLE')
+//     }
+//   }
+// }
+// 
+//
```
imuchnik, line 299: let's delete it?

jenkins-automation.groovy
```diff
@@ -172,6 +175,64 @@ behave -k -f=plain --logging-level=INFO --junit --junit-directory=test-results \
   }
 }
 
+
+// Requires: 
+// - BASE_URL: the starting url of your application
+//
+job('template-bdd-security'){
+  using 'template-base'
+
```
imuchnik, line 184: this is still using  templates? Should we just add a builder for it?

imuchnik:
Granted that it is the direction we want to continue in, small notes added.

Ooblioob:
yeah, I had mixed feelings about putting it in as a template, but until we have a focused vision for how we're doing builders I decided to just go with the old way.  Didn't want to hold us up from delivering something.  It shouldn't be too hard to refactor, so there's not a big risk.

jenkins-automation.groovy
Ooblioob, line 299: I hesitate to delete only because we don't necessarily know what we'd like to replace this with.  We have plans to replace the project builder / master builder with something that makes more sense and is less cumbersome, but until that is more flushed out I think it makes sense to keep it in there.  If nothing else, it serves as a reminder to tackle that problem and not just ignore it.

jenkins-automation.groovy
Ooblioob, line 184: Yes, I'd like us to for sure, but went with the old style since we haven't fully formed the idea.

imuchnik:
Going the "new" way, will create a builder for it.

OrlandoSoto:
Groovy :)

PR #10, opened 2015-10-08
Adding support for repo versions

imuchnik:
Dev-1249 in Jira for more info.

### DIFF
--- a/jobs/SampleBuildFlow.groovy
+++ b/jobs/SampleBuildFlow.groovy
@@ -1,7 +1,7 @@
 import FlowJobBuilder
 
 new FlowJobBuilder(
-        name: 'sample flow job',
+        name: 'GeneratedFlowJob',
         description: 'this our first stab at it',
         jobs:['job1', 'job2']
 ).build(this);
--- a/jobs/SampleBuildFlow.groovy
+++ b/jobs/SampleBuildFlow.groovy
@@ -1,0 +1,8 @@
+import FlowJobBuilder
+
+new FlowJobBuilder(
+        name: 'sample flow job',
+        description: 'this our first stab at it',
+        jobs:['job1', 'job2']
+).build(this);
+
--- a/jobs/SampleBuildFlow.groovy
+++ b/jobs/SampleBuildFlow.groovy
@@ -1,8 +1,15 @@
 import FlowJobBuilder
 
-new FlowJobBuilder(
+def oahMaster= new FlowJobBuilder(
         name: 'GeneratedFlowJob',
         description: 'this our first stab at it',
         jobs:['job1', 'job2']
 ).build(this);
+
+oahMaster.with{
+  logRotator{
+      numToKeep(365)
+  }
+
+}
 
--- a/jobs/SampleJsBuildJob.groovy
+++ b/jobs/SampleJsBuildJob.groovy
@@ -16,7 +16,6 @@ new JsJobBuilder(
         description: 'An example using a job builder for a Javascript build jobs project.',
         repos: repos,
         emails: developers,
-        use_versions: true,
-        logRotator: false
+        use_versions: true
 ).build(this)
 
--- a/jobs/SampleJsBuildJob.groovy
+++ b/jobs/SampleJsBuildJob.groovy
@@ -1,8 +1,10 @@
 import JsJobBuilder
+import FrontEndTestJobBuilder
 
 
 String basePath = 'JsJobSamples'
 List developers = ['irina.muchnik@cfpb.gov', 'daniel.davis@cfpb.gov']
+
 def repos = [
         [name: 'jenkins-automation', url: "https://github.com/cfpb/jenkins-automation@2.0"],
         [name: 'collab', url: "https://github.com/cfpb/jenkins-automation"]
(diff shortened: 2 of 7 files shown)


src/main/groovy/FrontEndTestJobBuilder.groovy
```diff
@@ -0,0 +1,33 @@
...
+    String name
+    String description
+    String gitBranch = 'master'
+    String pollScmSchedule = '@daily'
+    String tasks
+    String junitResults = '**/build/test-results/*.xml'
+    String artifacts = 'dist/'
+    List<String> emails
+    Boolean use_versions
+
```
Ooblioob, line 17: need to add a var for repos here

jobs/SampleBuildFlow.groovy
```diff
@@ -0,0 +1,15 @@
+import FlowJobBuilder
+
+def oahMaster= new FlowJobBuilder(
```
Ooblioob, line 3: we should rename this something more generic, since this is an example job.

jobs/SampleBuildFlow.groovy
```diff
@@ -0,0 +1,15 @@
...
+        name: 'GeneratedFlowJob',
+        description: 'this our first stab at it',
```
Ooblioob, line 5: Same here, put in some more generic text

jobs/SampleJsBuildJob.groovy
```diff
@@ -1,20 +1,26 @@
...
 
 String basePath = 'JsJobSamples'
 List developers = ['irina.muchnik@cfpb.gov', 'daniel.davis@cfpb.gov']
+
 def repos = [
-        [name: 'jenkins-automation', url: "https://github.com/cfpb/jenkins-automation"],
+        [name: 'jenkins-automation', url: "https://github.com/cfpb/jenkins-automation@2.0"],
         [name: 'collab', url: "https://github.com/cfpb/jenkins-automation"]
 ]
 folder(basePath) {
```
Ooblioob, line 12: folder is good, but our convention has been to use views more.  Can we change this to show using a view configured with a specific regex instead?

src/main/groovy/FrontEndTestJobBuilder.groovy
```diff
@@ -0,0 +1,42 @@
...
+        Job jsJob = new JsJobBuilder(
+                name: this.name,
+                description: this.description,
+                repos: this.repos,
+                emails: this.emails,
+                use_versions: true
+        ).build(factory)
+
+        print jsJob
```
Ooblioob, line 28: debugging?

src/main/groovy/FrontEndTestJobBuilder.groovy
```diff
@@ -0,0 +1,42 @@
+import JsJobBuilder
+import javaposse.jobdsl.dsl.DslFactory
+import javaposse.jobdsl.dsl.Job
+
+class FrontEndTestJobBuilder {
+
+    String name
+    String description
+    String gitBranch = 'master'
+    String pollScmSchedule = '@daily'
```
Ooblioob, line 10: Currently this field isn't being used, can we add it, either here or in the base?

src/main/groovy/FrontEndTestJobBuilder.groovy
```diff
@@ -0,0 +1,42 @@
...
+    String tasks
+    String junitResults = '**/build/test-results/*.xml'
+    String artifacts = 'dist/'
```
Ooblioob, line 13: ok, so a bunch of these aren't being used.  Delete or use them

src/main/groovy/JsJobBuilder.groovy
```diff
@@ -37,12 +49,14 @@ class JsJobBuilder {
...
             publishers {
                 archiveArtifacts artifacts
-                if (emails) {
-                    mailer emails.join(' ')
-                }
             }
+
         }
+
+        baseJob
```
Ooblioob, line 59: is this implicit return actually needed?  I had thought calling `baseJob.with` would be sufficient for what to return, but correct me if I'm wrong.

Ooblioob:
review completed, please address the items.

src/main/groovy/JsJobBuilder.groovy
imuchnik, line 59: @Ooblioob You are wrong. What is returned is the last call, which in this case would be the with() function, which is void.

src/main/groovy/FrontEndTestJobBuilder.groovy
imuchnik, line 28: Yep, you caught me

jobs/SampleJsBuildJob.groovy
imuchnik, line 12: Yes, we can. Can you create a jira issues we can throw these into?

src/main/groovy/JsJobBuilder.groovy
Ooblioob, line 59: Darn, ok.  I figured that might be the case.  Thanks for the explanation!

src/main/groovy/FrontEndTestJobBuilder.groovy
imuchnik, line 10: We have 2 options:
1. Delete them
2.  The job itself is pretty skimpy. We can functionality that leverages most of these fields. Will delete the rest then?

src/main/groovy/FrontEndTestJobBuilder.groovy
imuchnik, line 13: See above.

PR #12, opened 2015-10-29
Migrations

imuchnik:
Adding our testing utils for migrations


imuchnik:
@vickumar1981 please review

PR #13, opened 2015-11-06
remove files that should not be in source control

marcesher:


marcesher:
I munged a cleanup and an upgrade together. Note that this PR removes files we shouldn't have in SCM, and also upgrades to gradle 2.8 and groovy 2.4.4

If on a Mac, `brew update` and then `brew install gradle`

marcesher:
If you have an existing IDEA project, I've found the easiest way to deal with this PR is simply to create a new project from sources, choose the jenkins-automation project, and then refresh the gradle view to get the familiar gradle structure back

PR #16, opened 2015-11-14
Sitemonitor

marcesher:
This adds a basic SiteMonitor job, requiring the SiteMonitor plugin


src/main/groovy/jenkins/automation/utils/BaseJobBuilder.groovy
```diff
@@ -33,7 +33,11 @@ class BaseJobBuilder {
                 numToKeep(10)
             }
             publishers{
-                mailer emails.join(' ')  //TODO use extended email
+                allowBrokenBuildClaiming()
+
+                if(emails){
```
imuchnik, line 38: can pass an empty array? This guard statements is not enough. If it is in the base job, we should require or default it. 
Otherwise take it out of the base job.

src/main/groovy/jenkins/automation/utils/SiteMonitorJobBuilder.groovy
```diff
@@ -0,0 +1,48 @@
...
+
+        baseJob.with {
+            triggers {
+                if(cronSchedule) {
+                    cron(cronSchedule)
+                }
+            }
+
+            configure { project ->
+                project / publishers << 'hudson.plugins.sitemonitor.SiteMonitorRecorder' {
```
imuchnik, line 32: is having multiple SiteMonitorRecorder blocks? This will add without checking.

src/main/groovy/jenkins/automation/utils/BaseJobBuilder.groovy
marcesher, line 38: Groovy's truthiness will treat an empty array as null, thus this check is valid. Perfectly good reasons exist for not requiring email notifications on job failures

src/main/groovy/jenkins/automation/utils/SiteMonitorJobBuilder.groovy
marcesher, line 32: I don't understand

imuchnik, line 32: it will always append another SiteMonitor node and I am not sure that is the intent.

src/main/groovy/jenkins/automation/utils/BaseJobBuilder.groovy
imuchnik, line 38: You are absolutely right. I stand corrected.

PR #20, opened 2015-11-25
first pass at Checkmarx builder

OrlandoSoto:
First attempt at Checkmarx job builder using JAC


jobs/checkmarx-security-job-example.groovy
```diff
@@ -0,0 +1,14 @@
+import jenkins.automation.utils.CheckmarxSecurityJobBuilder
+
+def projectName ='foo'
+new CheckmarxSecurityJobBuilder(
+        name: "${projectName}checkmarx_security_job",
```
marcesher: let's use hyphens and not underscores, per our conventions for other jobs; let's also make sure there's a hyphen between projectname and the next stuff.

Also, "projectname-checkmarx-security-job" doesn't seem to match our conventions so far. Looking at our jenkins, it's simply "projectname-checkmarx"... any reason we wouldn't just stick with that?

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+            
+            scm {
+                git {
+                    remote {
+                        name('Repo to Scan')
+                        url(scanRepo)
+                    }
+                    
+                    branch('*/master')
+                    wipeOutWorkspace(true)
```
marcesher: is wiping out the workspace common?

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+    String highThreshold
+    String mediumThreshold
+    String lowThreshold
+
+    Job build(DslFactory factory){
+        factory.job(name){
+            it.description this.description
+            
+            scm {
+                git {
```
marcesher: seems to me we should probably support multiple repos, similar to how the base job builder does it. In fact, I wonder: shouldn't this whole thing use BaseJobBuilder; see JsJobBuilder for an example of using Base builder.

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+                        url(scanRepo)
+                    }
+                    
+                    branch('*/master')
+                    wipeOutWorkspace(true)
+                }
+            }
+
+            configure { project -> 
+                project / builders / 'com.checkmarx.jenkins.CxScanBuilder' (plugin:"checkmarx@7.2.3-31") {
```
marcesher: do we need to specify the version? seems onerous to maintain

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+                    'useOwnServerCredentials'('false')
+                    'serverUrl'('http://awe-codescan-w-d01')
```
marcesher: gah! let's get this hard-coded server name out of here and passed as a parameter

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+                    'projectName'('CheckmarxBuilder')
+                    'groupId'('ac43cb0d-034d-4b1e-9bf5-e7c1c46f71d2')  // Team
```
marcesher: groupid: parameterize this, I think

jobs/checkmarx-security-job-example.groovy
OrlandoSoto: agreed!

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
OrlandoSoto: no, it's not common in other Checkmarx jobs. I added wipeOutWorkspace to guarantee that Checkmarx always scans a clean copy of the repo

marcesher: there should be a less brute-force way to do this. check out the git options... I think there's one for simply reverting any changes. And really, I can't think of why a checkmarx job would ever have changes made to its workspace, so I'm really dubious that this would be necessary or would add value

marcesher: So much so, that I probably wouldn't add it to the builder as a default, and leave it up for people to add that behavior for their jobs if it actually became necessary, which I highly doubt it would

OrlandoSoto: I found this gem that could be a good optional parameter to have: https://jenkinsci.github.io/job-dsl-plugin/#path/job-scm-git-clean

marcesher: yeah, if you really do need to deal with artifacts lying around, that's the way to do it

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
OrlandoSoto: I removed the version and it ran without problems, @imuchnik any reason we should keep the plugin version?

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
OrlandoSoto: agreed!

imuchnik:
@OrlandoSoto please reabase and please add groovydoc comments for the builder and example to docs/examples.md

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
OrlandoSoto: ok

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
OrlandoSoto: @marcesher I changed this to use BaseJobBuilder,, as far as multi-repo support, we normally scan one repo per Jenkins job and generate one Checkmarx report. I could probably add my single repo code to ScmUtils and make it reusable? 
@Lamine any comments on this?

example-jobs/checkmarx-security-job-example.groovy
```diff
@@ -0,0 +1,18 @@
+import jenkins.automation.utils.CheckmarxSecurityJobBuilder
+
+def projectName ='foo'
+new CheckmarxSecurityJobBuilder(
+        name: "${projectName}-checkmarx",
+        description: "Sample checkmarx security job",
+        useOwnServerCredentials: false,
+        serverUrl: "http://awe-codescan-w-d01",
```
marcesher: let's not hard-code these serverUrl or groupId params; instead, we need to pass them in from the jobs; we can talk about making them environment variables or something else to abstract them, but they're internal/private data and consequently should not be in a public repo

OrlandoSoto: Environmental variables sound like a good idea here, since all jobs use the same server Url and Group ID's

src/main/groovy/jenkins/automation/utils/CheckmarxSecurityJobBuilder.groovy
```diff
@@ -0,0 +1,93 @@
...
+
+ *
+ * @param name  job name
+ * @param description  job description
+ * @param scanRepo  Github repo to scan
+ * @param checkmarxComment Additional comment(s) to include in the scan results
+ * @param useOwnServerCredentials If set to false then credentials from the Manage Jenkins page are used; serverUrl, username and password parameters are ignored. 
+ * If set to true then credentials must be specified in serverUrl, username and password parameters
+ *
+ * @see <a href="https://github.com/imuchnik/jenkins-automation/blob/gh-pages/docs/examples.md#bdd-security-job-builder" target="_blank">TODO Checkmarx job Example</a>
```
imuchnik: the links should point to cfpb  forks not imuchnik

OrlandoSoto: ok

PR #24, opened 2015-12-09
updated bdd security builder

OrlandoSoto:
Added additional parameters to the job builder


imuchnik:
Please add an example in docs/examples.md

PR #25, opened 2015-12-10
Splitting out into proper namespace and adding Env Utils

imuchnik:
The rest is just noise and details.


marcesher:
I manually merged this and added a few small things. We do have some cleanup to do, but I wanted to move this along so that we can start using it

PR #27, opened 2015-12-14
Modify flow builder

vickumar1981:
- Allow string value to flow builder for jobs run in parallel
- Allow artifacts to be null (optional) / causes certain jobs to fail if specified.


sleitner:
:+1:

marcesher:
:+1:  but lets update the example jobs to show usage. Also, make sure the groovydocs are updated so that usage shows up in the API viewer

PR #31, opened 2015-12-17
fix broken examples; add new example for BuildFlow with a workspace

marcesher:


example-jobs/SampleBuildFlow-example.groovy
```diff
@@ -1,15 +1,31 @@
...
+        jobs:['job1', 'job2']
+).build(this)
+
+exampleFlowWithWorkspace.with{
+    scm {
+        git('https://github.com/cfpb/qu.git', 'master')
+    }
+
+    configure { node ->
+        node << buildNeedsWorkspace(true)
```
imuchnik, line 46: Why do we need that?

example-jobs/envTestJob.groovy
```diff
@@ -2,7 +2,7 @@
 import static jenkins.automation.utils.EnvironmentUtils.isDev
 
 
-def env=${ENVIRONMENT}
+def env="${ENVIRONMENT}"
```
imuchnik: Not sure that does anything?

marcesher: It makes it work

example-jobs/SampleBuildFlow-example.groovy
marcesher, line 46: Because flow jobs that pull from a repo need to have a workspace. Without configuring it to have a workspace, you get nasty errors in jenkins when the job runs. I've been thinking that we might even want to just set this by default... there's no harm to it when you don't need it, and when you do need it, unless you know exactly what you're missing, the error is not clear

example-jobs/envTestJob.groovy
```diff
@@ -2,7 +2,7 @@
...
 if (isDev(env)){
```
vickumar1981: I guess I didn't see this before, but why do we pass in the environment to the isDev() method?

Can't isDev() check if ${ENVIRONMENT} == 'dev'?

If anything, I feel like the env variable should be encapsulated within the  class EnvironmentUtils { }

https://github.com/cfpb/jenkins-automation/blob/master/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy

example-jobs/checkmarx-security-job-example.groovy
```diff
@@ -8,7 +8,7 @@ new CheckmarxSecurityJobBuilder(
         serverUrl: "http://sample-serverUrl",
         username: "sample-username",
         password: "sample-password",
-        groupId: "${groupId}",
+        groupId: "",
         scanRepo: "https://github.com/OrlandoSoto/ckan-browser-tests",
```
OrlandoSoto, line 13: hmmm why did we take this out?

example-jobs/envTestJob.groovy
marcesher: Because $ENVIRONMENT is only available inside job scripts; it's not accessible directly inside of jenkins-automation classes.

The way around that is to make EnvironmentUtils a proper class with a constructor that takes a string, and stores that string as state, and then turns those existing static methods into nonstatic class methods that simply read the internal state:
psuedo:

```
def env = ""
def EnvironmentUtils(String environment) {
  env = environment
}

def getEnv() {
  return env.ToUpperCase()
}
def isDev() {
  return getEnv() == "DEV"
}

def isProd(){
  return getEnv() == "PROD"
}
```

and then in scripts:

```
envUtil = new EnvironmentUtils("${ENVIRONMENT}")

...

if( envUtil.isDev()){
...
}
```

just a thought.

imuchnik: @marcesher You are right.
@vickumar1981 long story about groovy script vs. class context and bindings, but in short: "No. We can't."  We can chat more about it if you interested in the mechanics.

example-jobs/checkmarx-security-job-example.groovy
marcesher, line 13: because the variable is not defined in the example file, and consequently it errors. It's not taking it out of the builder itself, just the example.

An alternate fix would be to define groupId in the example job, but I don't think that adds anything.

imuchnik, line 13: I fixed that  in my PR

vickumar1981:
@marcesher @imuchnik  Can we close this PR out?  It's been up for 3 weeks and needs to be rebased if we're going to merge it.

marcesher:
@imuchnik @vickumar1981 I'll review this and push a mergeable version

marcesher:
@imuchnik @vickumar1981 This PR is now fixed

vickumar1981:
:+1:

example-jobs/SampleBuildFlow-example.groovy
imuchnik, line 46: if that is the case, we should make it a default. I was thinking of expanding our scmUtils. I can work it into it then.

imuchnik:
Looks good! Thanks for the test.

example-jobs/SampleBuildFlow-example.groovy
marcesher, line 46: @imuchnik I agree it should be a default. Though it doesn't have anything to do with scmUtils.

Do you want to add this configure block into BuildFlow.groovy and take it out of the example? If not, I can do that. Let me know

PR #34, opened 2015-12-18
Improve description of jobFlow and modify behavior

sleitner:
Simplify the JobFlow api a little by making it possible to use both jobs and jobFlow.


vickumar1981:
+1 LGTM.

PR #36, opened 2015-12-21
Environ utils

vickumar1981:
@marcesher @sleitner  @imuchnik 

Changes to environ utils as discussed on previous PR #31.

env is now a member variable of the EnvironUtils class.  The isDev() and isProd() functions are non-static.

### DIFF
--- a/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
@@ -14,7 +14,18 @@ class EnvironmentUtils {
  *  relies on ENVIRONMENT env variable in Jenkins
  */
 
+    private static final EnvironmentUtils instance = new EnvironmentUtils()
     String env
+
+    private EnvironmentUtils() {
+        // do nothing
+    }
+
+    public static EnvironmentUtils getInstance(String env) {
+        instance.env = env
+        return instance()
+    }
+
 
     boolean isDev() {
         return getEnv() as Environment== Environment.dev
@@ -27,6 +38,7 @@ class EnvironmentUtils {
     boolean isStage() {
         return getEnv() as Environment == Environment.stage
     }
+
 
     String getEnv() {
         try {
--- a/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
@@ -17,18 +17,21 @@ class EnvironmentUtils {
     String env
 
     boolean isDev() {
-        return env.toLowerCase() as Environment== Environment.dev
+        return getEnv() as Environment== Environment.dev
     }
 
     boolean isProd() {
-        return env.toLowerCase() as Environment == Environment.prod
+        return getEnv() as Environment == Environment.prod
+    }
+
+    boolean isStage() {
+        return getEnv() as Environment == Environment.stage
     }
 
     String getEnv() {
         try {
             return env.toLowerCase() as Environment
         } catch (all) {
-
             return all.getLocalizedMessage()
         }
     }
--- a/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
@@ -23,7 +23,7 @@ class EnvironmentUtils {
 
     public static EnvironmentUtils getInstance(String env) {
         instance.env = env
-        return instance()
+        return instance
     }
 
 
--- a/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
@@ -17,11 +17,11 @@ class EnvironmentUtils {
     String env
 
     boolean isDev() {
-        return env.toUpperCase() as Environment== Environment.DEV
+        return env.toLowerCase() as Environment== Environment.dev
     }
 
     boolean isProd() {
-        return env.toUpperCase() as Environment == Environment.PROD
+        return env.toLowerCase() as Environment == Environment.prod
     }
 
     String getEnv() {
--- a/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
@@ -7,24 +7,26 @@ package jenkins.automation.utils
  */
 
 
- class EnvironmentUtils {
+class EnvironmentUtils {
 
 /**
  *  Enum for Jenkins environments
  *  relies on ENVIRONMENT env variable in Jenkins
  */
 
-    static isDev(String env) {
+    String env
+
+    boolean isDev() {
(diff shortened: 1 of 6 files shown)


src/main/groovy/jenkins/automation/utils/EnvironmentUtils.groovy
```diff
@@ -7,24 +7,26 @@ package jenkins.automation.utils
...
  *  relies on ENVIRONMENT env variable in Jenkins
  */
 
-    static isDev(String env) {
-        return env.toUpperCase() as Environment== Environment.DEV
+    String env
+
+    boolean isDev() {
+        return env.toLowerCase() as Environment== Environment.dev
     }
```
sleitner, line 33: isStage?

vickumar1981, line 33: doesn't exist.  let me add it.

imuchnik, line 33: Those were arbitrarily names. We will need to still talk about it.

sleitner:
:+1:

imuchnik:
:-1: Making the class non-static means that you have to instantiate it in your script and now the dev need to know about obscure groovy script bindings? I do not understand what do we win by this?

vickumar1981:
The other option, the way it is, requires you to pass the environment into the isDev() and isProd() functions.  Monadic functions are easier to understand, IMO.

imuchnik:
This requires passing the variable on class creation. I do not see a strong argument for it . Either way, the variable is not available inside the class. I think that this requires more work for the developers, which kind of defeats the purpose. IF everyone feels very strongly about it, we can spend more time to figure out whether we can source that file into the scripts(as a script, rather than a class). That would achieve both goals.

sleitner:
I think either way, it's prrretty clear what is going on, and either way, the user will have to look at the same groovy or (more likely) example code. I prefer this solution slightly because you pass fewer arguments and it's marginally clearer what each call is doing as a result (even though you have to make one extra call).

If we're only sourcing one variable from a script, would we really be gaining anything over just instantiating the class with $ENVIRONMENT?

imuchnik:
@sleitner We pass exactly the same amount of arguments. We  gain just that- not having instantiate a class and still having it available. This has a cross-cutting nature, similarly to LOG function.

vickumar1981:
@imuchnik   I don't see how it's different than JsJobBuilder or FlowBuilder where you have to pass arguments into the class creation.

def myJob = new JsJobBuilder(
        name: jobName,
        description: 'OAH front-end build',
        emails: ...

Why not just have a static method in JsJobBuilder called build job which takes all these variables as parameters?

So you'd have this:

class JsJobBuilder {
    ...
   static Job build(DslFactory factory, String name, String description, String email ...){
        ...
    }
}

EnvironmentUtils takes one parameter on construction right now, but that's why it's good b/c it's extensible.  In fact, static methods can be avoided altogether with a singleton design pattern :)

vickumar1981:
So logger() is a good example.  That's a singleton usually.  In python, in java...always a singleton.

sleitner:
"We pass exactly the same amount of arguments"

How is that? You pass your environment on every `isX` call. And it's very clear what instantiating a class means, whereas it's not as clear what passing an argument will do.

imuchnik:
> In fact, static methods can be avoided altogether if you use a singleton design pattern :)

@vickumar1981  I don't see us enforcing a singleton pattern anywhere and we definitely do not need more then one instance of this class, which is also not true for builders. 

> So logger() is a good example. That's a singleton usually. In python, in java...always a singleton.

Note the absence of **new** keyword.

imuchnik:
Also, where is the usage example and tests?

vickumar1981:
you wouldn't need more than one instance of a builder either, if you pass all the arguments directly into the function...

build(DslFactory factory, String name, String description, String email ...){

Then I could do:

def job1 = JsJobBuilder.build (...)
def job2 = JsJobBuilder.build (...)

the problem is that arguments might get added on later, and that having a function call with 10 parameters is confusing.  So, we encapsulate them in a class.  I see the same thing with EnvironUtils, where we might want to add more settings later.  If EnvironUtils had 10 arguments instead of 1, then we wouldn't be making the case for using a static method.

imuchnik:
@vickumar1981 I disagree,  we do need multiple instances of builders. 

I would argue against making this class do other things because we would be violating a "single responsibility principle". All this class should ever do is answer one question:"What environment am I in?"

vickumar1981:
@imuchnik I can provide a decent use case - plugins.  If the list of plugins is different on prod vs. dev vs. staging, we might need to know about that.  A job could then, for example, ask EnvironUtils if that plugin is available for that environment.  

So I agree there's a single responsibility for the class, but it might, at some point later on, answer more than just a single question.  we might end up asking it a few more things about the environment down the road...about plugins and environment specific paths and stuff.

imuchnik:
@vickumar1981 Sorry, I still do not see it.  Environment determination is a runtime state.  Plugin availability is an orthogonal concern that is not dependent on the environment, i.e this is something only Jenkins knows in any environment.

sleitner:
@imuchnik How is plugin availability orthogonal to environment? Each environment has a Jenkins and each Jenkins has a set of plugins. They're perfectly correlated. We should probably unify the jenkinses, but for now, that's a fine example. And you can come up with any number of others: isDevIP(ip) for example.

That said, I think we're straying from the point a little, which is really that all of the static methods need to maintain a state, and we're asking the user to pass that state in over and over again, rather than maintaining it internally.

imuchnik:
@sleitner orthogonal in way that plugins are not dependent on environments and vice versa. In simpler terms, knowing the environment does not guarantee the presence of the plugins and environments do not know anything about plugins present.   IsDevIp() will have live in some other utils.

Agreed about this conversation going off on the tangent. 

Perhaps seeing an example on code will help me see your thought process better?

vickumar1981:
code just looks like this:

def env = new EnvironmentUtils (env: "${DEPLOY_ENV}")

and then:

if (env.isDev()) {
   ...//do something
}

if (env.isProd()) {
  ...//do something else
}

imuchnik:
My problem is with new(). I don't think devs should have to do that in the script. 
Also that is not singleton pattern at all. If you are really opposed to the static class, but want a singleton we should do something like logger, where you would do Logger.getLogger()  to get a handle on the class instance.

vickumar1981:
Yes.  I agree with that.  That's why I pointed out the singleton pattern.  IMO, EnvironmentUtils() should be a singleton.

Want me to make EnvironUtils a singleton() ?  Since that's something we all agree upon.

imuchnik:
Yep, that would solve my issue with having developers new() up an instance .

vickumar1981:
Done.  new way to invoke is like this:

def env = EnvironmentUtils.getInstance('dev')

sleitner:
I'll leave it to Irina to plus one, but it's worth adding a usage example or comment.

imuchnik:
:+1: but I am  not merging anything until @sleitner  does the same ;)

Also add example and this probably broke the test.

vickumar1981:
fixed the broken test.

sleitner:
woot! :+1:

PR #42, opened 2016-01-04
Add PeriodicJobBuilder

m3brown:
PeriodicJobBuilder essentially creates cron jobs in Jenkins.
- cron syntax: `H/15 * * * *` runs every 15 minutes
- optional repo to clone if you need to run a script from a repo

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -44,6 +44,7 @@ import jenkins.automation.builders.BddSecurityJobBuilder
 
 ```
 
+
 ## Base Job Builder
 
 
@@ -127,6 +128,34 @@ new JsJobBuilder(
 ).build(this)
 
 ```
+
+## Periodic Job Builder
+
+```
+import jenkins.automation.builders.PeriodicJobBuilder
+
+List developers = ['irina.muchnik@cfpb.gov', 'daniel.davis@cfpb.gov']
+String projectUrl = 'https://github.com/cfpb/bash-automation'
+String cronSchedule = 'H/10 * * * *'
+
+def periodicJob = new PeriodicJobBuilder(
+        name: 'periodic-job-name',
+        description: 'An example using a job builder for a periodic task.',
+        emails: developers,
+        cronSchedule: cronSchedule,
+        repo: projectUrl
+).build(this)
+
+periodicJob.with {
+  steps {
+    shell('''
+./execute_script_from_projectUrl_repo.sh
+''')
+  }
+}
+
+```
+
 ## Using MultiScm Utility 
 
 ```
--- a/src/main/groovy/jenkins/automation/builders/PeriodicJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/PeriodicJobBuilder.groovy
@@ -1,0 +1,59 @@
+package jenkins.automation.builders
+
+import javaposse.jobdsl.dsl.DslFactory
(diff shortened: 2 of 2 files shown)


marcesher:
While I appreciate the work here, I really struggle with the value this adds. Basically, this simplifies adding a cron schedule, but I don't think merits a whole new builder.

here's what I believe about Builders, in principle:

1) Prefer sticking with job-dsl vs new abstractions; the bar should be pretty high for adding new builders
2) we should not provide simple abstractions over top of already simple abstractions, and job-dsl is often quite simple. 

The question I have is: does it really add value to add a simple wrapper around adding a cron schedule? I'd much rather that people using job-dsl know how to do that without relying on a builder abstraction.

I know this is fuzzy and the guidelines on when should we add a builder vs. not aren't clear, and it's definitely up for debate.

Here are a few things to consider to guide decisions?
- Does the builder abstract a gnarly configure block (such as sitemonitor builder)? If so... probably worth it
- Does the builder create a custom "job" type, such as the BuildFlowBuilder? If so... probably worth it
- Does the builder represent a job that is used so consistently within the org that further modification of that job is not likely (such as the checkmarx builder)? If so, probably worth it.
- Does the builder simply add a wrapper around non-custom / non-configure-block job-dsl bits? If so, probably not worth it
- Does the builder expose something that saves a line or two of code, but makes it annoying to provide further customization (often the case with `scm{}` blocks)? If so, probably not worth it

Bottom line: I'd bias toward using job-dsl as much as possible, even if it means _more_ code, over additional abstractions.

When to break this rule? I'd say on a per-project basis, when you look at existing jobs and realize that many of them are highly consistent except for 1 or 2 things that differ, such as the Qu "task" jobs. In that case, it's appropriate to create a custom builder _inside that project_ and use that

m3brown:
I understand your concern about adding a bunch of negligibly incremental builders to the builder stack, and I'm more than happy to do all the logic in my project's groovy code.  I was just trying to follow the workflow that I had observed with our existing stack (i.e. the SiteMonitorBuilder).

Having said that, we have a lot of jobs that run on a schedule, whether it's `H/15 * * * *` or `@hourly` or `@daily`.  At a quick glance it appears that at least 8 Qu jobs have a schedule.  Does this Builder save a ton of time over using BaseJobBuilder with a cron blerb? Not really, but it is something we'd have to do over and over again for many if not all of our projects.

If we expect a high bar for entry as a JobBuilder, maybe we should detail our expectations (basically, what you said above) somewhere in the readme or the documentation, I can't find anything on this topic.

marcesher:
@m3brown These are all excellent comments

I heartily agree about documenting this guidance. I'll add a Jira issue for it.

Re: lots of qu jobs, I think this is likely where it gets into that grayer area, which we can document guidance on to the best of our ability. With Qu, not only did those jobs have cron schedules, but they also had damn-near-identical shell blocks, log-parser-publisher blocks, and mask-password-parameter blocks. Those several things combined led me to create a QuTaskBuilder for use with those task-ish jobs (https://omg-private-url/marcesher/qu-jenkins-automation/blob/master/src/main/groovy/QuTaskBuilder.groovy)

imuchnik:
@marcesher @m3brown I am with Marc. The rule of a thumb for a builder- it should produce a viable job. I doubt we have jobs that have a schedule, but do nothing else. We can add it as an optional parameter.

PR #43, opened 2016-01-07
Add shallow clone and disable submodule options for scm utils

sleitner:


imuchnik:
@sleitner either you are missing an example or I am not sure this will work

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
```diff
@@ -24,8 +32,9 @@ class ScmUtils {
                         branch("*/tags/$version")
                     }
                     relativeTargetDir(repo.sub_directory)
+                    shallowClone(shallow)
```
imuchnik: shallowClone(repo.shallow) should be all you you need here

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
```diff
@@ -9,6 +9,7 @@ class ScmUtils {
 
 
     static void project_repos(context, repos, use_versions = true) {
+        Boolean shallow
```
imuchnik: initialize it here.

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
```diff
@@ -16,6 +17,13 @@ class ScmUtils {
                 def parsed_url = parsed_out_url[0]
 
                 def version = parsed_out_url[1]
+
+                if(repo.containsKey('shallow')) {
```
imuchnik: This puzzles me. An example of what you have in mind will be helpful.

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
sleitner: repo.shallow will be null if the shallow key isn't provided, and shallowClone needs a boolean argument.

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
sleitner: added one to example.md

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
imuchnik: Initialize it to false (default) above, but I think when it is null it is defaulted to false by the plugin anyways.
http://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.scm.GitContext.shallowClone

sleitner: Is this what you mean? 

```
        Boolean shallow = false
        shallow = repo.shallow
```

If so, then shallow is null and then `shallowClone(shallow)` receives a null argument and throws an error.

imuchnik: you can use Elvis operator and do something like
shallowClone(repo.shallow?:false)
and that should do it. 

?: ==> elvis ( one of my groovy favs)

sleitner: ah okay :+1:, one line then.

sleitner:
should this be the default? why do we ever need the history of a repo to be deployed via jenkins?

vickumar1981:
:+1:

Scotchester:
I agree, shallow clone should be the default.

PR #44, opened 2016-01-08
Make shallow clone the default

sleitner:


src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
```diff
@@ -28,7 +28,8 @@ class ScmUtils {
                         branch("*/tags/$version")
                     }
                     relativeTargetDir(repo.sub_directory)
-                    shallowClone(repo.shallow?:false)
+                    //shallow clone is the default
+                    shallowClone(repo.shallow!=false?true:false)
```
imuchnik, line 32: This is too much. The original code  shallowClone(repo.shallow?:false) should do the trick
What elvis does is check if expression is true and assign itself to a value, otherwise some other value. 
So if repo.shallow is true, shallowClone is true, otherwise false.
Groovy evaluates 0 and null as false, so we should be ok.

imuchnik:
Example and groovydoc comment for the new parameters? What is disable_submodule variable? Adding groovydoc comment helps with understanding and also adds it to docs.

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
sleitner, line 32: null _should_ evaluate to true here if shallow clone is to be the default. (I try to avoid ternaries like this because the intent can get confusing.)

Scotchester:
@sleitner and @imuchnik Can we resolve this? Would be nice to have in place.

imuchnik:
I am going to close this and add it in another PR along with everything that is needed for docs.

sleitner:
:+1: to @imuchnik

imuchnik:
Turns out it is already in the code.

Scotchester:
@imuchnik I'm confused. We want shallow clone to be the default. It seems like the default is still `false`.

PR #45, opened 2016-01-08
Refactor adding job defaults out of BaseJob and into Utils

imuchnik:
That makes it independently available to other builders.

### DIFF
--- a/src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
@@ -2,6 +2,7 @@ package jenkins.automation.builders
 
 import javaposse.jobdsl.dsl.DslFactory
 import javaposse.jobdsl.dsl.Job
+import jenkins.automation.utils.CommonUtils
 
 /**
  * The very first and basic building block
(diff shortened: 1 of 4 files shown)


vickumar1981:
:+1:

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
```diff
@@ -0,0 +1,22 @@
...
+ * @param context  delegate passed in context\
+ */
+class CommonUtils {
+    static void add_defaults(context) {
+        context.with {
+            wrappers {
+                colorizeOutput()
+            }
+            logRotator {
+                numToKeep(10)
```
marcesher: I think 10 is too low. Let's change the default to something more reasonable and in line with how a lot of our jobs are currently configured.... I'd say at least 30

src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
```diff
@@ -25,46 +26,18 @@ class BaseJobBuilder {
...
-            }
-            logRotator {
-                //TODO: make it parametrized to builder
-                numToKeep(10)
-            }
-            publishers{
-                allowBrokenBuildClaiming()
-
-                if(emails){
+            CommonUtils.add_defaults(delegate)
```
marcesher: please change this to `addDefaults`, to match the style of the rest of the code

imuchnik: should we allow overrides?

marcesher: Well... that means that every single builder will also need to support that parameter. Is it worth it, compared with just adding a new block in your jobs.groovy for that job in the cases where you do need to overwrite it?

imuchnik: we can leave it with a sensible default and optional ability to override. Net total effect is the same with configure block, but they achieve it with a one-liner parameter.

marcesher:
@imuchnik thanks!

Issue #53, opened 2016-01-24
Convert email into a reusable, parameterized configure block


marcesher:
We can make the editable email config more useful by turning it into a reusable, parameterized configure block.

The BaseJobBuilder and friends can then use it, as can any other jobs that don't use the builders.

suggested signature something like:

```
    static Closure editableEmail(List<String> triggers = ["Failure", "Fixed"], List<String> emails=[], sendToDevelopers=false, includeCulprits=false, sendToRecipientList=true) {

    }
```

The purpose here is to make it really easy to do the thing that we need to do most often with email, which is send an email with default subject, default body, a configurable recipient list, and only to recipients (sending to developers and culprits is often a bad idea because we don't want internal jenkins emails going to external contributors, who might show up in the developers or culprits list)

Jobs that need more configuration can still easily use editable email directly without this helper, which is why it's important we keep this block simple and not load it up with all manner of configurability

Issue #51, opened 2016-01-23
Checkmarx example job failing


marcesher:
When I set up a job to run the example jobs, the checkmarx job fails with:

```
ERROR: (ScmUtils.groovy, line 15) No such property: url for class: java.lang.String
```

PR #54, opened 2016-01-25
Addresses #53

imuchnik:
Adds extendedEmail block


src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
```diff
@@ -19,4 +24,31 @@ class CommonUtils {
...
+ * @param triggers List<String> triggers E.g Failure, Fixed etc...
+ * @param String emails Comma separated string of emails
+ * @param sendToDevelopers Default false,
+ * @param includeCulprits Default false,
+ * @param sendToRecipientList Default true
+ *
+ * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#common-utils" target="_blank">BDD job Example</a>
+ */
+
+    static void extendedEmail(context, List<String> triggers = ["Failure", "Fixed"], String emails, sendToDevelopers = false, includeCulprits = false, sendToRecipientList = true) {
```
marcesher: Should emails be a `List<String>` instead, to retain compatibility with the Builders? I don't have a strong opinion, but it seems to me that we should remain consistent

docs/examples.md
```diff
@@ -179,3 +179,28 @@ job('test') {
...
+ job('example'){
+    CommonUtils.addDefaults(delegate)
+    }
+```
+### Extended Email
+
+import jenkins.automation.utils.CommonUtils
+
+ job('example'){
+   extendedEmail(delegate,
```
marcesher: This looks wrong. Wouldn't it be CommonUtils.extendedEmail(....)?

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
imuchnik: The extendedEmail block takes a string http://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.PublisherContext.extendedEmail and I am a little hesitant to diverge from that. I can still make it work for the builder doing cool groovy things http://grails.asia/groovy-list-to-string inside the builders themselves. **OR** I can override the function to also take in a list with a signature like so: +    static void extendedEmail(context, List<String> triggers = ["Failure", "Fixed"], List emails, sendToDevelopers = false, includeCulprits = false, sendToRecipientList = true)

docs/examples.md
imuchnik: Oh why, yes! Yes, indeed!

imuchnik:
@marcesher comments addressed.

example-jobs/checkmarx-security-job-example.groovy
```diff
@@ -10,7 +10,7 @@ new CheckmarxSecurityJobBuilder(
         username: "sample-username",
         password: "sample-password",
         groupId: "",
-        scanRepo: "https://github.com/OrlandoSoto/ckan-browser-tests",
+        scanRepo:[[url: "https://github.com/OrlandoSoto/ckan-browser-tests"]],
```
marcesher, line 13: What's this? Looks like an unrelated change, and an incorrect one

imuchnik, line 13: No, it is correct. The builder was changed to accept a list of repos to scan. See #51.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
marcesher: ah, well then. Here are my 2 cents:

I think we should change the API to support a string, and only a string.

Yes, there will be some work for us to change any existing configs, but it's a few minutes of work and probably not a big deal.

I'll merge this version in now, and then we can plan a time -- probably next week -- with all interested parties to make the cutover from the List to a string.

I think we should be trying to write a) as little code as possible, b) code that doesn't deviate from job-dsl for no good reason

example-jobs/checkmarx-security-job-example.groovy
marcesher, line 13: ah, I see, this is for #51 

thanks.

PR #55, opened 2016-01-25
Change signature for addExtendedEmail 

marcesher:
This keeps the required args up front and optional args thereafter


imuchnik:
:+1:

PR #56, opened 2016-01-26
removed double quotes from sed command

OrlandoSoto:


OrlandoSoto:
@marcesher here's the change we discussed earlier

marcesher:
Thanks!

Issue #50, opened 2016-01-23
BuildFlowJob: default to requiring a workspace


marcesher:
create all build flow jobs with a workspace. Near as I can tell, there's no downside, all upside

marcesher:
accomplished with:

```
buildFlowJob('example') {
    buildNeedsWorkspace()
}
```

Though this does require `v1.42` of the job-dsl-plugin to be installed

marcesher:
hmmm...

when I updated my jenkins to the latest job-dsl, configure blocks started failing. Let's hold off on this.

marcesher:
ok, figured it out. The failing configure block is the one in the example job with buildNeedsWorkspace. It fails because that method now exists as of 1.42. I'll update the example job to show how to do it in earlier and the most recent version of job-dsl

imuchnik:
plugin dependency bumped to 1.42

PR #57, opened 2016-01-27
forces all build flow jobs to have a workspace. 

marcesher:
Addresses #50


marcesher:
Why do this?

Any build flow job that does SCM polling, or anything else that requires the build flow job to have state, requires a workspace. Problem is, by not setting that, the behavior you get from jenkins does not clearly indicate this is the problem.

By creating all build flow jobs with a workspace, we avoid that nonsense entirely. For jobs that don't technically require it, there's no downside that I can tell.

imuchnik:
Watch out! EVERYTHING WILL REBUILD NOW!!!!

PR #58, opened 2016-02-01
Add 'Unstable' as default email trigger; clean up extended email example

marcesher:
If you want emails on failure, you quite likely also want emails on unstable, since both likely require action

For email examples, I removed the bits that were just duplicating the defaults.


vickumar1981:
:+1: LGTM

PR #59, opened 2016-02-02
Make sure only master branch is watched

imuchnik:
ScmUtils multiScm block was watching all branches by default, if no tag was specified vis "use_versions" variable.

### DIFF
--- a/src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
@@ -17,7 +17,7 @@ class JsJobBuilder {
         fingerprint()
         defaultExcludes()
     }
-    List<String> emails
+    def emails
     Boolean use_versions
 
     def repos = [];
--- a/src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
@@ -27,6 +27,9 @@ class ScmUtils {
                     if (use_versions && version!=null) {
                         branch("*/tags/$version")
                     }
+                    else {
+                        branch master
+                    }
                     relativeTargetDir(repo.sub_directory)
                     shallowClone(repo.shallow?:false)
                     if(disable_submodule){
--- a/src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
@@ -3,7 +3,7 @@ package jenkins.automation.utils
 /**
  * Utility class to provide nicer, terser DSL for common tasks
  *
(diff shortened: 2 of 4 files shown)


imuchnik:
@marcesher @Scotchester @vickumar1981 Please review.

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
```diff
@@ -27,6 +27,9 @@ class ScmUtils {
                     if (use_versions && version!=null) {
                         branch("*/tags/$version")
                     }
+                    else {
+                        branch master
```
marcesher: does this need to be quoted?

imuchnik: Probably does :wink:  http://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.scm.GitContext.branch.

Good catch!

src/main/groovy/jenkins/automation/utils/ScmUtils.groovy
```diff
@@ -27,6 +27,9 @@ class ScmUtils {
                     if (use_versions && version!=null) {
                         branch("*/tags/$version")
                     }
+                    else {
+                        branch "master"
```
Scotchester, line 37: Why does the above `branch` statement use parentheses, but this one doesn't?

imuchnik, line 37: groovy "poetic mode". You  can drop the parenthesis if there is only one argument.

imuchnik:
Added some docs for shallow clone.

imuchnik:
hmm, that is a delegate. Somewhat special for groovy. It sets the closure execution context. It may only confuse people, if I added to the docs.

Scotchester:
Fair enough.

src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
```diff
@@ -17,7 +17,7 @@ class JsJobBuilder {
         fingerprint()
         defaultExcludes()
     }
-    List<String> emails
+    def emails
```
Scotchester, line 20: Can someone with more Groovy experience than me vet this change?

imuchnik, line 20: We want to be able to pass either a list or string for emails. This makes it dynamic variable. It will use "duck-typing" at runtime.

Scotchester:
Thanks for fixing the shallow clone default! :+1:

src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
Scotchester, line 20: Good enough for me.

PR #63, opened 2016-02-04
Cleanup, docs and making pollSchedule optional

imuchnik:
Housekeeping a bit..

### DIFF
--- a/src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
@@ -4,14 +4,35 @@ import javaposse.jobdsl.dsl.DslFactory
 import javaposse.jobdsl.dsl.Job
 import jenkins.automation.utils.ScmUtils
 
+
+/**
+ * @param name          Job name
+ * @param description   Job description
+ * @param pollSchedule  Optional
+ * @param artifacts     Optional Closure defaulted to closure below :
+                                  {
+                                     pattern("dist/")
+                                     fingerprint()
+                                     defaultExcludes()
+
+                                     }
+
+ * @param emails  List or String of notification email addresses
+ * @param repos List of repos to watch
+ * @param use_versions same as BaseJobBuilder
+ * <p>
+ *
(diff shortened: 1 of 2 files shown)


src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
```diff
@@ -4,14 +4,35 @@ import javaposse.jobdsl.dsl.DslFactory
 import javaposse.jobdsl.dsl.Job
 import jenkins.automation.utils.ScmUtils
 
+
+/**
+ * @param name          Job name
+ * @param description   Job description
+ * @param pollSchedule  Optional
```
Scotchester: Typo: should be `pollScmSchedule`

Scotchester: Should also have an actual explanation for what it does :)

src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
```diff
@@ -4,14 +4,35 @@ import javaposse.jobdsl.dsl.DslFactory
...
+                                  {
+                                     pattern("dist/")
+                                     fingerprint()
+                                     defaultExcludes()
+
+                                     }
+
+ * @param emails  List or String of notification email addresses
+ * @param repos List of repos to watch
+ * @param use_versions same as BaseJobBuilder
```
Scotchester, line 22: It'd be nicer if this just said what it was instead of making me reference a different file.

src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
imuchnik: Agreed. Well said. Write it  and it will also show up here: http://cfpb.github.io/jenkins-automation/

src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
imuchnik, line 22: Sure. The documentation can have a bit more meat on it. But checkout this: http://cfpb.github.io/jenkins-automation/.

PR #65, opened 2016-02-05
Making groovydoc more descriptive

imuchnik:


src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
```diff
@@ -4,8 +4,31 @@ import javaposse.jobdsl.dsl.DslFactory
...
+                                  {
+                                     pattern("dist/")
+                                     fingerprint()
+                                     defaultExcludes()
+
+                                     }
+
+ * @param emails  List or String of notification email addresses
+ * @param repos List of repos to watch
+ * @param use_versions flag to check out the repo at a specific tag. The tag is parsed out from url property appended after '@' sign.
```
Scotchester: This one should probably make note of the fact that it only applies in multi-SCM situations.

imuchnik: Yes. Done.

Scotchester:
:+1: Thank you!

PR #66, opened 2016-02-05
Parametrizing nodejs version

imuchnik:


src/main/groovy/jenkins/automation/builders/JsJobBuilder.groovy
```diff
@@ -17,6 +17,7 @@ import jenkins.automation.utils.ScmUtils
                                      }
 
  * @param emails  List or String of notification email addresses
+ * @param node_version  Version of nodeJs installation to use. Relies on nodejs plugin
```
Scotchester: I'd like to elaborate here that it must correspond to one of the Node installation names set up in Jenkins. Currently, it could be seen as just "Give us the version number of Node you want to run and you got it!"

Scotchester:
:+1:

PR #67, opened 2016-02-08
changed Failure threshold to 0

OrlandoSoto:
changed Failure threshold to 0, it was previously 1
This will allow the test to fail the Jenkins job if the number of failed tests is > 0


OrlandoSoto:
@marcesher @lfatty

marcesher:
Thanks a lot, @OrlandoSoto

PR #69, opened 2016-02-19
Add new utils; clean up docs

marcesher:
Adds two new utils and cleans up docs and some (not all) example code


docs/examples.md
```diff
@@ -31,128 +42,94 @@ new CheckmarxSecurityJobBuilder(
...
-            """
-    ).build(this);
-
-    customFlowJob.with{
-      logRotator{
-          numToKeep(365)
-      }
-
-    }
+FlowJobBuilder(
```
Scotchester, line 61: This needs a `new` in front of it, doesn't it?

docs/examples.md
```diff
@@ -31,128 +42,94 @@ new CheckmarxSecurityJobBuilder(
...
+        name: 'sample-flow-job',
+        description: 'A sample build flow job that builds a list of jobs in order',
+        jobs:['job1', 'job2']
+).build(this).with {
+  logRotator{
+      numToKeep(365)
```
Scotchester, line 67: Is there a reason to encourage overriding our default of `numToKeep(30)` here and in the example below?

docs/examples.md
```diff
@@ -31,128 +42,94 @@ new CheckmarxSecurityJobBuilder(
...
 import jenkins.automation.builders.JsJobBuilder
 
-
 String basePath = 'JsJobSamples'
 List developers = ['irina.muchnik@cfpb.gov', 'daniel.davis@cfpb.gov']
 
 def repos = [
         [name: 'jenkins-automation', url: "https://github.com/cfpb/jenkins-automation@2.0"],
-        [name: 'collab', url: "https://github.com/cfpb/jenkins-automation", shallow: true]
+        [name: 'collab', url: "https://github.com/cfpb/collab", shallow: true]
```
Scotchester, line 99: `shallow: true` is the default now; this could be removed.

docs/examples.md
```diff
@@ -31,128 +42,94 @@ new CheckmarxSecurityJobBuilder(
...
         [name: 'other', url: "https://github.com/cfpb/jenkins-automation", disable_submodule: true]
```
Scotchester, line 100: What does the `name` in a `repos` item do?

docs/examples.md
marcesher, line 61: good catch

docs/examples.md
marcesher, line 67: I think it's useful to at least show how... I don't think that's exactly encouraging it

docs/examples.md
marcesher, line 99: will do

docs/examples.md
marcesher, line 100: good question. Nothing, near as I can tell

docs/examples.md
```diff
@@ -31,128 +42,94 @@ new CheckmarxSecurityJobBuilder(
...
+        [name: 'collab', url: "https://github.com/cfpb/collab", shallow: true]
+]
+new BaseJobBuilder(
+        name: "sample-job-with-multiscm",
+        description: "A sample with multiple source control repositories",
+).build(this).with {
     multiscm {
-        ScmUtils.project_repos(delegate, this.repos, use_versions)
-        }
+        ScmUtils.project_repos(delegate, repos, true)
```
Scotchester, line 130: Our typical pattern is to [pass `repos` into the constructor](https://[GHE]/CFPB/cfgov-jenkins-automation/blob/master/jobs/cfgov-deploy-apache-conf.groovy#L16), which elimintates the need for the `multiscm` block in the `with`.

Should we do that in this example?

example-jobs/SampleBaseJobs.groovy
```diff
@@ -0,0 +1,39 @@
...
+def repos = [
+        [name: 'jenkins-automation', url: "https://github.com/cfpb/jenkins-automation@2.0"],
+        [name: 'collab', url: "https://github.com/cfpb/collab", shallow: true]
+]
+new BaseJobBuilder(
+        name: "sample-base-job-with-multiscm",
+        description: "A sample with multiple source control repositories",
+).build(this).with {
+    multiscm {
+        ScmUtils.project_repos(delegate, repos, true)
```
Scotchester, line 37: Same comment about passing in `repos`

docs/examples.md
marcesher, line 130: BaseJob does not support repos (nor, I think, should it)

docs/examples.md
```diff
@@ -196,15 +174,40 @@ job('example'){
...
-         
+    CommonUtils.addExtendedEmail(delegate, emails = 'foo@example.com, bar@example.com') 
 }
    
-   
 // override accepts emails as a list. Compatible with builders
 job('example'){
-  CommonUtils.addExtendedEmail(delegate, 
-  emails = ['someperson@email.com', 'someotherperson@email.com']) 
+    CommonUtils.addExtendedEmail(delegate, emails = ['foo@example.com', 'bar@example.com'])
```
Scotchester, line 182: Similar to the question about `repos`, we have the `emails` abstraction that can be added in to the constructor. Should we document that pattern instead?

docs/examples.md
Scotchester, line 130: Ah, right you are.

docs/examples.md
Scotchester, line 182: NM, per the fact that baseJobBuilder doesn't support `emails`, either.

Scotchester, line 182: Gah, spoke too soon. It DOES have `emails` support. So I reopen this line of inquiry.

PR #71, opened 2016-02-19
Adding Failure scan plugin support to defaults

imuchnik:
we use everywhere and that will prevent from config being re-written


marcesher:
Looks good to me. Thanks, @imuchnik

Scotchester:
Late-breaking thought I had on this: What happens if a third party picks up our jenkins-automation stuff but isn't using a plugin that our code assumes, like the Build Failure Analyzer?

PR #74, opened 2016-02-19
add a performance publisher util

marcesher:
This is one I'm not quite sure about.

On one hand, it's quite useful. It exposes the common things that I think we want to tweak in our load test jobs (based on my experience writing our load test jobs).

On the other hand, it doesn't expose _all_ configurable items.

My thinking here is that it's a good enough, useful block for our load tests.

For jobs that require full control over the performance publisher, they're free to use this as a guide but directly configure it in their jobs.

### DIFF
--- a/example-jobs/SampleBaseJobs.groovy
+++ b/example-jobs/SampleBaseJobs.groovy
@@ -39,6 +39,16 @@ new BaseJobBuilder(
     }
 }
 
+new BaseJobBuilder(
+        name: "sample-base-job-with-performance-publisher",
+        description: "A job with a performance publisher. It does not include the actual bits that run the load tests"
+).build(this).with {
+    steps {
+        shell("echo 'Run jmeter tests here'")
+    }
+    CommonUtils.addPerformancePublisher(delegate,failedThresholdPositive=10, failedThresholdNegative=10, unstableThresholdPositive=5, unstableThresholdNegative=5)
+}
+
 def repos = [
         [url: "https://github.com/cfpb/jenkins-automation@2.0"],
         [url: "https://github.com/cfpb/collab"]
--- a/src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
+++ b/src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
@@ -106,6 +106,39 @@ class CommonUtils {
     }
 
     /**
+     * Utility to add a performance publisher block, for use in performance tests
+     *
+     * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#common-utils" target="_blank">Common utils</a>
+     */
+
+    static void addPerformancePublisher(context, String reportPattern="**/results/*.jtl", String unstableResponseTimeThreshold="", int failedThresholdPositive, int failedThresholdNegative, int unstableThresholdPositive, int unstableThresholdNegative) {
+        context.with {
+            configure {
+                it / publishers << 'hudson.plugins.performance.PerformancePublisher' {
+                    errorFailedThreshold 2
+                    errorUnstableThreshold 1
+                    errorUnstableResponseTimeThreshold unstableResponseTimeThreshold
+                    relativeFailedThresholdPositive failedThresholdPositive
+                    relativeFailedThresholdNegative failedThresholdNegative
+                    relativeUnstableThresholdPositive unstableThresholdPositive
+                    relativeUnstableThresholdNegative unstableThresholdNegative
+                    modeRelativeThresholds false
+                    configType 'ART'
+                    modeOfThreshold true
+                    compareBuildPrevious true
+                    modePerformancePerTestCase true
+                    modeThroughput true
+                    parsers {
+                        'hudson.plugins.performance.JMeterParser' {
+                            glob reportPattern
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    /**
      * Common string for creating and activating a python 2.7 virtualenv in a shell block
      *
      * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#common-utils" target="_blank">Common utils</a>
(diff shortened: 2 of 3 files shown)


example-jobs/SampleBaseJobs.groovy
```diff
@@ -39,6 +39,16 @@ new BaseJobBuilder(
...
 }
 
+new BaseJobBuilder(
+        name: "sample-base-job-with-performance-publisher",
+        description: "A job with a performance publisher. It does not include the actual bits that run the load tests"
+).build(this).with {
+    steps {
+        shell("echo 'Run jmeter tests here'")
+    }
+    CommonUtils.addPerformancePublisher(delegate,failedThresholdPositive=10, failedThresholdNegative=10, unstableThresholdPositive=5, unstableThresholdNegative=5)
```
vickumar1981, line 49: we should default these parameters for the thresholds in the CommonUtils.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
```diff
@@ -106,6 +106,39 @@ class CommonUtils {
...
+                it / publishers << 'hudson.plugins.performance.PerformancePublisher' {
+                    errorFailedThreshold 2
+                    errorUnstableThreshold 1
+                    errorUnstableResponseTimeThreshold unstableResponseTimeThreshold
+                    relativeFailedThresholdPositive failedThresholdPositive
+                    relativeFailedThresholdNegative failedThresholdNegative
+                    relativeUnstableThresholdPositive unstableThresholdPositive
+                    relativeUnstableThresholdNegative unstableThresholdNegative
+                    modeRelativeThresholds false
+                    configType 'ART'
```
vickumar1981, line 126: do we ever use a different configType here?  If so, this might be something we want the option to override/set.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
```diff
@@ -106,6 +106,39 @@ class CommonUtils {
...
     /**
+     * Utility to add a performance publisher block, for use in performance tests
+     *
+     * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#common-utils" target="_blank">Common utils</a>
+     */
+
+    static void addPerformancePublisher(context, String reportPattern="**/results/*.jtl", String unstableResponseTimeThreshold="", int failedThresholdPositive, int failedThresholdNegative, int unstableThresholdPositive, int unstableThresholdNegative) {
+        context.with {
+            configure {
+                it / publishers << 'hudson.plugins.performance.PerformancePublisher' {
```
vickumar1981, line 117: Is the convention here to use '<<' or '/'?  I think this will always append a new performance publisher node in the xml.  I'm not sure what happens if there are two such blocks, but it might be better to use '/' here.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
```diff
@@ -106,6 +106,39 @@ class CommonUtils {
     }
 
     /**
+     * Utility to add a performance publisher block, for use in performance tests
+     *
+     * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#common-utils" target="_blank">Common utils</a>
+     */
+
+    static void addPerformancePublisher(context, String reportPattern="**/results/*.jtl", String unstableResponseTimeThreshold="", int failedThresholdPositive, int failedThresholdNegative, int unstableThresholdPositive, int unstableThresholdNegative) {
```
vickumar1981, line 114: With this many parameters to pass into the function call, does it make sense to factor out into a separate class?  Or to make them attributes of CommonUtil perhaps?

http://www.informit.com/articles/article.aspx?p=1375308

Quoting Robert Martin's Clean Code: "The ideal number of arguments for a function is zero (niladic). Next comes one (monadic), followed closely by two (dyadic). Three arguments (triadic) should be avoided where possible. More than three (polyadic) requires very special justification -- and then shouldn't be used anyway."

imuchnik, line 114: Some of these are defaulted. I think it is ok to have as a function, not a class.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
imuchnik, line 117: we actually might want multiple blocks here, but it is a good point. I am not sure.

imuchnik:
Since it is an additive change, I think there is no harm in merging. 
@marcesher  thank you for remembering to add the groovydoc comments and examples.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
vickumar1981, line 114: It's just less clear that's all.  Every time you use the function, you'll have to go back and look it up, like what do I pass in again?  It's more of a suggestion though.  Currently, you're passing in 7 parameters, and let's just say you want to add more at a later time.  The function approach is less extendable.

imuchnik, line 114: Yes. Fair point.

example-jobs/SampleBaseJobs.groovy
marcesher, line 49: I went back and forth on that. Having built a number of these load test jobs, I can't think of any common enough settings that would be good defaults.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
marcesher, line 114: This is part of why I'm not sure it's worth even including this thing.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
marcesher, line 117: Good catch, Vic.

src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
marcesher, line 126: I've never seen us use a different config type

PR #75, opened 2016-02-26
Bringing the Salesforce builder to the masses

imuchnik:
This need not be private. Open sourcing...

@marcesher @Scotchester @OrlandoSoto @vickumar1981 One or more of you kind gentleman, please review.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -118,7 +118,7 @@ new JsJobBuilder(
 
 import jenkins.automation.builders.SalesforceAntJobBuilder
 
-List developers = ['irina.muchnik@cfpb.gov', 'daniel.davis@cfpb.gov']
+List developers = ['irina@mailinator.gov', 'yolo@examle.io']
 
 def repo = "https://github.com/cfpb/saleforce-automation"
 
@@ -127,7 +127,7 @@ new SalesforceAntJobBuilder(
         description: 'An example using a job builder for a Salesforce Ant JobBuilder build jobs project.',
         repoUrl: repo,
         emails: developers,
-        antTaskName: "retrieveUnpackaged",
+        antTasks: ["retrieveUnpackaged"],
         antInstallerName:"ant-latest"
 ).build(this)
 
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -128,7 +128,29 @@ new JsJobBuilder(
 ).build(this)
 
 ```
-## Using MultiScm Utility 
+# Salesforce Build Job
+
+
+```
+
+import jenkins.automation.builders.SalesforceAntJobBuilder
+
+List developers = ['irina.muchnik@cfpb.gov', 'daniel.davis@cfpb.gov']
+
+def repo = "https://github.com/cfpb/saleforce-automation"
+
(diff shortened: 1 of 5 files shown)


docs/examples.md
```diff
@@ -112,7 +112,30 @@ new JsJobBuilder(
...
 ```
 
-## Using MultiScm Utility 
+# Salesforce Build Job
+
+```
+
+import jenkins.automation.builders.SalesforceAntJobBuilder
+
+List developers = ['irina.muchnik@cfpb.gov', 'daniel.davis@cfpb.gov']
```
marcesher: let's use example.com emails in all our examples / docs

src/main/groovy/jenkins/automation/builders/SalesforceAntJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+
+        def job = new BaseJobBuilder(
+                name: name,
+                description: description,
+                emails: emails
+        ).build(factory);
+
+        job.with {
+            logRotator {
+                numToKeep(365)
```
marcesher: Why the override?

src/main/groovy/jenkins/automation/builders/SalesforceAntJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+            }
+            scm {
+                git {
+                    remote {
+                        url(repoUrl)
```
marcesher, line 48: Do we know it'll always be a single URL? Seems like this should be wrapped in an if statement, to allow for flexibility, i.e. if folks do have multiple repos, they can skip repoUrl and use multiscm directly

src/main/groovy/jenkins/automation/builders/SalesforceAntJobBuilder.groovy
```diff
@@ -0,0 +1,63 @@
...
+                    }
+                }
+            }
+            steps {
+                ant {
+                    target(antTaskName)
```
marcesher: recommend using `targets` instead

src/main/groovy/jenkins/automation/builders/SalesforceAntJobBuilder.groovy
imuchnik: Good catch. Will remove.

docs/examples.md
imuchnik: Sure. No objections.

src/main/groovy/jenkins/automation/builders/SalesforceAntJobBuilder.groovy
imuchnik, line 48: Trying with one right now. Will change if needed later. Not sure if SF supports development across multiple repos.

src/main/groovy/jenkins/automation/builders/SalesforceAntJobBuilder.groovy
imuchnik: done.

PR #76, opened 2016-03-14
Rename sample jobs to use _ instead of -

vickumar1981:
@imuchnik @marcesher @m3brown  - addresses https://jira.demo.cfpb.gov/browse/DEV-1565


marcesher:
Hey @vickumar1981 @imuchnik 

I'm late to this, but let's revisit it. We have some jobs that are all lowercase with underscores, some with CamelCase and underscores, some with "Sample" and "example", some with just Sample, some with example, some with neither.

I don't have any strong opinions on a format, other that it be:

1) consistent
2) not redundant (i.e. not have "Sample" and "Example")

Thanks

PR #78, opened 2016-04-27
Script Runner and example job renaming

marcesher:
1. Adds slightly modified versions of RestApiScriptRunner and RestApiJobManagement from
   https://github.com/sheehan/job-dsl-gradle-example
2. renames all example jobs for consistency and absence of redundancy

Details:
1. Script runners:
   Irina had originally included these in the internal
   jenkins-automation-private repo, for use with recovering the "Mother
   Seed Job".
   I think this is so useful for a local development workflow that I'm
   bringing it into jenkins-automation
   It will then be usable by any repo that uses jenkins-automation.
   The idea is to use this as part of a local development workflow for
   job-dsl jobs that mitigates the slow-and-annoying git workflow we have
   ot use now
2. Example job renaming

These jobs are already in an `example-jobs` folder.
I see no need to have them have `example` or `sample` in the name.
In addition, they should be named consistently, not the island of misfit
names that they were prior to this commit.


src/main/groovy/jenkins/automation/rest/RestApiJobManagement.groovy
```diff
@@ -0,0 +1,123 @@
...
+class RestApiJobManagement extends MockJobManagement {
+
+    final RESTClient restClient
+
+    RestApiJobManagement(String baseUrl) {
+        if (!baseUrl.endsWith("/")) {
+            baseUrl += "/"
+        }
+        restClient = new RESTClient(baseUrl)
+        restClient.handler.failure = { it }
```
vccabral, line 19: What does the it in the block do?

src/main/groovy/jenkins/automation/rest/RestApiJobManagement.groovy
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Override
+    String getConfig(String jobName) throws JobConfigurationNotFoundException {
+        String xml = fetchExistingXml(jobName)
+        if (!xml) {
+            throw new JobConfigurationNotFoundException(jobName)
+        }
+
+        xml
```
vccabral, line 33: I'm starting to like groovy.

src/main/groovy/jenkins/automation/rest/RestApiJobManagement.groovy
marcesher, line 19: which lines specifically, Chris?

src/main/groovy/jenkins/automation/rest/RestApiJobManagement.groovy
marcesher, line 33: 😃

src/main/groovy/jenkins/automation/rest/RestApiJobManagement.groovy
vccabral, line 19: { it }

m3brown, line 19: > For closures accepting a single argument, the keyword it can be used to represent that single parameter rather than explicitly specifying the single parameter's name

http://www.javaworld.com/article/2074349/core-java/groovy-s-special-words.html

marcesher: I'm not sure.

m3brown:
What I'd like to see is some documentation (either in the PR, or preferably in readme markdown) on how to use the script runner.

If the solution is a link to a document elsewhere that's fine, but it also could be a code snippet that i can copy and paste into my dev environment.

marcesher:
@m3brown I'll definitely add that, but I am not certain of the particulars yet because I need to get this PR merged first so that I can test using it from within a JAC repo. 

Ultimately, though, from any JAC repo you should be able to do this:

`./gradlew rest -Dpattern=<pattern> -DbaseUrl=<baseUrl> [-Dusername=<username>] [-Dpassword=<password>]`
- `pattern` - ant-style path pattern of files to include
- `baseUrl` - base URL of Jenkins server
- `username` - Jenkins username, if secured
- `password` - Jenkins password or token, if secured

imuchnik:
👍 Should also be added to the starter repo

vickumar1981:
I think if you loaded those parameters from a config file, it might be better, and also serve as documentation.  That is, allow the settings to be defaulted from a file, but overridden from the command line.

I don't particularly like the pattern argument either.  Instead of loading files from a directory, we should be loading classes from a package using inspection.  So if I said something like '--package=jenkins.jobs', then any class under that package would go ahead and be run.

marcesher:
We can debate the particulars later. I need to merge this so that I can see if it even works from within a separate repo

marcesher:
Sweet. This works exactly as expected. I'll add docs to here and the starter repo soon (tracking this work in Jira).

@vickumar1981 I like the config file idea. However, I don't think the package thing applies, because jobs don't live in packages, just folders. The pattern arg lets you target specific job files:

`./gradlew rest -Dpattern=jobs/environment_test_job.groovy`, for example, which during development is probably what you're looking for

vickumar1981:
yeah, i understand what the pattern does.  I just think it's a limitation.  All your files have to be inside the same directory (what if I want different folders for my jobs?  does it handle recursion?  is there a switch for this?)  

I could have multiple classpaths, for example, with jobs in different projects, but logically grouped under the same package.  What if I just send you a JAR file with all my jobs already compiled, for example? The latter method allows users of JAC greater flexibility.  Like in the environment_test_job.groovy example above, I just think it should be inside a package, and I don't care where that package is, so long as it's able to be located inside $CLASSPATH.

marcesher:
@vickumar1981 I think you are way overthinking this, tbh

marcesher:
@vickumar1981 Plus, ant pattern matcher is extremely flexible. I don't know if it'd support your JAR example, but that's not a thing I think would actually exist in real life anyway. And remember, this is a development-time tool, designed to help you iterate on jobs before pushing them

vickumar1981:
Sure, but I don't see why we're limiting it to run only files/scripts vs a more elegant and flexibile approach of loading the class from the classpath.

```
task rest(dependsOn: 'classes', type: JavaExec) {
   main = 'jenkins.automation.rest.RestApiScriptRunner'
    ...
}
```

Why does gradle itself want you to specify a class instead of a file/directory for example, as to where the Main method is located?  I get that this is a single resource and not a pattern, but...

When you specify dependencies, when you declare a managed, or unmanaged resource, in basically every configuration option for gradle, sbt, maven, ant, you're specifying a package, b/c packages are how the JVM logically groups classes.  src/main/[groovy | scala | java] is automatically loaded into the CLASSPATH by convention, but we still allow a configurable option to modify/change that convention.

I'm just saying that if we follow what the JVM build tools do, then we won't be tying ourselves down as much.

marcesher:
@vickumar1981 I think this is beyond my JVM knowledge, and we should probably take it to pairing

PR #80, opened 2016-05-06
Move extended email to new format for DSL v1.43

m3brown:
In Jenkins 1.43, the DSL format for extended email changed.

The CommonUtils.addExtendedEmail function was updated to reflect the new format.

Side Effects:
- the `triggers` parameter was renamed to `triggerList`
- the trigger values now need to be lowercase (e.g. 'failure' instead of 'Failure')

Change Description: https://github.com/jenkinsci/job-dsl-plugin/wiki/Migration#extended-email
API Docs: https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.PublisherContext.extendedEmail


m3brown:
We should hold off merging this until all affected teams are aware of the change.

OrlandoSoto:
Good idea @m3brown !

PR #92, opened 2016-10-21
Add preSendScript argument to addExtendedEmail method

fna:
### Review

@vickumar1981

### DIFF
--- a/src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
@@ -15,7 +15,6 @@ import jenkins.automation.utils.CommonUtils
  * @param name used to name the job
  * @param description job description
  * @param emails list of developer to get notifications
- * @param preSendScript a groovy script to modify default email before sending it
  * <p>
  *
  * @see <a href="https://github.com/imuchnik/jenkins-automation/blob/gh-pages/docs/examples.md#base-job-job-builder"
@@ -27,7 +26,6 @@ class BaseJobBuilder {
     String name
     String description
     List<String> emails
-    String preSendScript
 
     Job build(DslFactory factory) {
         factory.job(name) {
@@ -35,14 +33,8 @@ class BaseJobBuilder {
             CommonUtils.addDefaults(delegate)
             publishers {
                 if (emails) {
-                    if (preSendScript) {
-                        publishers {
-                            CommonUtils.addExtendedEmail(delegate, emails, false, true, false, true, preSendScript)
-                        }
-                    } else {
-                        publishers {
-                            CommonUtils.addExtendedEmail(delegate, emails)
-                        }
+                    publishers {
+                        CommonUtils.addExtendedEmail(delegate, emails)
                     }
                 }
             }
--- a/src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
@@ -15,6 +15,7 @@ import jenkins.automation.utils.CommonUtils
  * @param name used to name the job
  * @param description job description
  * @param emails list of developer to get notifications
+ * @param preSendScript a groovy script to modify default email before sending it
  * <p>
  *
  * @see <a href="https://github.com/imuchnik/jenkins-automation/blob/gh-pages/docs/examples.md#base-job-job-builder"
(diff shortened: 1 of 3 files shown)


src/main/groovy/jenkins/automation/utils/CommonUtils.groovy
```diff
@@ -42,8 +42,8 @@ class CommonUtils {
      * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#common-utils" target="_blank">Common utils</a>
      */
 
-    static void addExtendedEmail(context, List<String> emails, List<String> triggerList = ["failure", "unstable", "fixed"], sendToDevelopers = false,  sendToRequester = true, includeCulprits = false, sendToRecipientList = true) {
-        addExtendedEmail(context, emails.join(","), triggerList, sendToDevelopers, sendToRequester, includeCulprits, sendToRecipientList)
+    static void addExtendedEmail(context, List<String> emails, List<String> triggerList = ["failure", "unstable", "fixed"], sendToDevelopers = false,  sendToRequester = true, includeCulprits = false, sendToRecipientList = true, preSendScript = "$DEFAULT_PRESEND_SCRIPT", postSendScript = "$DEFAULT_POSTSEND_SCRIPT") {
```
vickumar1981: @imuchnik   what should these be defaulted to?  null?  empty string?

@fna i think you have to add a setting to the BaseJobBuilder to pass in the presend and postsend arguments to CommonUtil.addExtendedEmail function.

Here's what he's trying to add to the email: https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.ExtendedEmailContext.preSendScript

m3brown:
For the sake of argument, this is adding a fairly special case into the framework.  I have three opinions on this:

1) For the BaseJobBuilder, I think `preSendScript` is out of scope.  We've gone back and forth on what should be in the BaseJobBuilder several times and we've settled on keeping it simple.  For example, we don't have flags for the other variables in the email function (triggerList, sendToDevelopers, sendToRequester, includeCulprits, sendToRecipientList), why should preSendScript get special treatment?

2) Similar to # 1, I don't like how preSendScript is the only email feature enabled by the addExtendedEmail function.  Why not be complete and add support for the other variables:
- $DEFAULT_RECIPIENTS
- $DEFAULT_REPLYTO
- $DEFAULT_CONTENT
- $DEFAULT_POSTSEND_SCRIPT
- $DEFAULT_SUBJECT

3) If a job in jenkins-automation (I prefer not BaseJobBuilder, but maybe a new Builder called CustomEmailJobBuilder that has flags for overriding options) adds preSendScript (and more!) functionality, we need to make sure the documentation reflects the new functionality.
- [Documentation for BaseJobBuilder](https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#base-job-builder) is out of date in this proposed PR
- [Documentation for addExtendedEmail](https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#extended-email) is also out of date

src/main/groovy/jenkins/automation/builders/BaseJobBuilder.groovy
```diff
@@ -26,15 +27,22 @@ class BaseJobBuilder {
...
 
     Job build(DslFactory factory) {
         factory.job(name) {
             it.description this.description
             CommonUtils.addDefaults(delegate)
             publishers {
                 if (emails) {
-                    publishers {
-                        CommonUtils.addExtendedEmail(delegate, emails)
+                    if (preSendScript) {
```
imuchnik: I do not see a need for this in the job builder. You already have it in the addExtendedEmail method and it is defaulted. I would just pass it in. It will evaluate to null and default if not provided.

imuchnik:
@fna I agree with @m3brown. The reason support for extended email resides in Utils, so that you can add it to jobs easily. We purposely kept it job-type agnostic.

fna:
@imuchnik @m3brown just so I understand it correctly
1. Remove the changes to the BaseJobBuilder
2. Update documentation for addExtendedEmail 

Could you confirm that?

m3brown:
That works for me.

imuchnik:
Yes, that is correct.

imuchnik:
@fna Please also add an example here: https://github.com/cfpb/jenkins-automation/blob/master/docs/examples.md#extended-email

imuchnik (approved):
Looks good!

imuchnik:
@m3brown @vickumar1981 any last comments before I merge?

PR #93, opened 2016-10-27
Update docs for addExtendedEmail

fna:
## Review

@imuchnik 

Please note that the gh-pages branch has actual code in it (https://github.com/cfpb/jenkins-automation/tree/gh-pages/src), @m3brown thinks it should be removed and that's why I haven't updated the code in the branch.


imuchnik:
@fna @m3brown The docs in gh-pages are auto generated. I will remove the code, but there is not need to PR to this branch.

PR #95, opened 2017-01-18
Updating to latest jobs-dsl-core plugin v1.57

jls5177:
In the latest release the runDslEngine() method was deprecated and removed. This
commit makes the required API changes.


imuchnik:
Thank you @jls5177 !

PR #98, opened 2017-03-01
Support UTF-8 charset.

keitap:
Supports UTF-8 charset so when you use CJK characters in DSL code, it won't get garbled.


keitap:
ping @imuchnik . Let me know if you need anything else.

imuchnik:
Thank you @keitap! Sorry for the delay.  👍

PR #107, opened 2017-09-28
WIP - pipeline job builder for discussion

acrewdson:
very WIP

### DIFF
--- a/src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
@@ -1,0 +1,89 @@
+package jenkins.automation.builders
+
+import javaposse.jobdsl.dsl.DslFactory
+import javaposse.jobdsl.dsl.Job
+import jenkins.automation.utils.CommonUtils
+
+
+/**
+ * The basic PipelineJob builder
+ *
+ * @param stages list of maps describing the stages; see example linked below for the shape of the map
+ * @param name job name
+ * @param description job description
+ * @param pipelineScript optional string containing the raw pipeline script (use instead of `stages`)
+ * @param pollScmSchedule optional string in cron format to trigger builds on a scheduled interval
+ * @param emails list of email addresses to receive notifications
+ * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#TODOXXXXXXXXXXXX" target="_blank">Pipeline job builder example</a>
+ */
+
+class PipelineJobBuilder {
+    List<Map> stages
+    String name
+    String description
+    String pipelineScript = ""
+    String pollScmSchedule
+    List<String> emails
+
+    /**
+     * @param DSL factory class, provided by Jenkins when executed from build context
+     * @return pipeline job
+     */
+
+    Job build(DslFactory factory) {
+        factory.pipelineJob(name) {
+
+            def warningText = ''
+            if (stages && pipelineScript) {
+                warningText += """
+                    ansiColor('xterm') {
+                        echo '\\033[31m! `pipelineScript` param has no effect because ' +
+                        '`stages` was also supplied; choose one or the other\\033[0m'
+                    }
+                """
+            }
+
+            it.description this.description
+
+            CommonUtils.addDefaults(delegate)
+
+            publishers {
+                if (emails) {
+                    publishers {
+                        CommonUtils.addExtendedEmail(delegate, emails)
+                    }
+                }
+            }
+
+            if (pollScmSchedule) {
+                triggers {
+                    scm pollScmSchedule
+                }
+            }
+
+            if (stages) {
+                def stagesStr = stages.collect {
+                    """
+                        stage("${it.stageName}") {
+                            build job: "${it.jobName}"
+                            ${it.parameters ? 'parameters: ' + it.parameters : ''}
+                        }
+                    """
+                }.join("\n")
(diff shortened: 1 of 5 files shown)


src/main/groovy/jenkins/automation/rest/RestApiScriptRunner.groovy
```diff
@@ -3,7 +3,6 @@
 package jenkins.automation.rest
 
 import javaposse.jobdsl.dsl.DslScriptLoader
-import javaposse.jobdsl.dsl.ScriptRequest
```
acrewdson, line 6: @imuchnik might know how this one snuck in here -- was it an unused import? If so I will break it out into its own commit.

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * @param DSL factory class, provided by Jenkins when executed from build context
+     * @return pipeline job
+     */
+
+    Job build(DslFactory factory) {
+        factory.pipelineJob(name) {
+
+            def warningText = ''
+            if (stages && pipelineScript) {
```
acrewdson, line 37: I'm not a huge fan of this from an API design perspective. I think me and @imuchnik are open to ideas for making this less confusing.

marcesher: My 2 cents: don't support stages or pipelinescript. I know... crazy talk! Hear me out: the value in these builders IMO is a) consistent approach to bootstrapping a job, b) sane defaults. But the "guts" of a job, IMO, aren't the bread and butter of these builders. I say leave out the guts and just enable users to do something like:

```
new PipelineBuilder(name: "my-pipeline", description="blah blah", whatevs="ok").build(this).with {
   

    agent {
        label 'my-agent-node'
    }

    options {
        skipDefaultCheckout()
    }

    stages {
        stage('Init') {
            steps {
                script { 
                    @Library('myGlobalLibrary') 
                    def stuff = new thingie()
                }
                sh 'pwd'
            }
        }

}
```

That way, the job still flows nicely from top to bottom and is easily readable and gets the 2 benefits of builders that I described above.

My advice: start simple.

marcesher: More: I don't think we'd be doing anyone any favors by abstracting away a tiny bit of boilerplate by supporting an array of stages, for example. IMO it's better for the entire team's learning to **not** abstract, but to encourage people to learn how to use pipeline syntax without hand-holding. That makes it much easier for teams to switch to `Jenkinsfile` if that becomes desirable, as well

yunghuffy: I have to say I support this idea a lot. When I was sweating and trying to implement this, I came up with a sample job that looked like this:

```groovy
import java.io.File
import jenkins.automation.builders.PipelineJobBuilderDSL

String scriptFileText = new File('test_pipeline.groovy').text

def pipelineJob = new PipelineJobBuilderDSL(
        name: 'Pipeline builder',
        description: 'this is a simple pipeline job',
        pipelineScript: scriptFileText,
        emails: ['jane@example.com', 'joe@example.com']
).build(this);
```

And the test_pipeline.groovy was just declarative pipeline syntax

```
pipeline {
    agent {
        label 'platform-ops'
    }

    options {
        ansiColor('xterm')
        //buildDiscarder(logRotator(numToKeepStr:'10'))
        skipDefaultCheckout()
        //timestamps()
    }

    stages {
        stage('Init') {
            steps {
                sh 'env | sort'
            }
        }

        stage('Build and Test') {
            steps {
              echo "I'm doing things"
            }
        }

        stage('Deploy') {
            steps {
              echo "I'm deploying things"
            }
        }
    }
}
```

That really, really simplifies the implementation in the `Builder` class to boil down to something like the following, and it allows us to alleviate some of @acrewdson 's concerns about putting a big if clause in a class that changes functionality.

```
definition {
               cps {
                   sandbox(false)
                   script(pipelineScript)
               }
           }
```

I have a branch that is also very much a WIP at https://github.com/cfpb/jenkins-automation/compare/master...champain:jacob_pipeline-job-builder that might add to this discussion.

acrewdson: I like where @champain is going with this. I think we ended up with the code in this PR because we were assuming we needed to try to match the `FlowJobBuilder` API, since we'd be porting jobs over from that. Based now on @marcesher 's feedback, it sounds like we're freed from that constraint, which is great.

marcesher: When I look at the build flow jobs I've created using the FlowJobBuilder, they all use the string arg and not the array arg

acrewdson: @marcesher ah, good to know -- and it makes sense

imuchnik: @marcesher Agree passing in the script is straightforward and simple. The builder supports just passing the script as a parameter. I felt it was good to have list of stage as well for now, since this was a piece of functionality that existed in FlowJobBuilder. If it turns out, it has little or no benefit, will delete it.

acrewdson:
closing so that the non-WIP PR can be opened soon

PR #108, opened 2017-10-02
DEV-2295 Pipeline job builder

imuchnik:
Adding support for Jenkins pipelines

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -86,6 +86,59 @@ new FlowJobBuilder(
 }
 ```
 
+## Pipeline Builder
+
+```groovy
+import jenkins.automation.builders.PipelineJobBuilder
+
+def script = """
+             node {
+                stage('First stage') {
+                    build job: 'Job 1',
+                    parameters: [
+                        [\$class: 'StringParameterValue', name: 'foo', value: 'bar'],
+                    ]
+                }
+            }
+"""
+
+def pipelineJob = new PipelineJobBuilder(
+    name: 'Pipeline builder',
+    description: 'This is a simple pipeline job',
+    pipelineScript: script,
+    emails: ['jane@example.com', 'joe@example.com']
+
+).build(this);
+
+pipelineJob.with {
+    logRotator {
+        numToKeep(365)
+    }
+}
+
+def pipelineJobStages = new PipelineJobBuilder(
+        name: 'Pipeline builder',
+        description: 'this is a simple pipeline job',
+        stages: [[
+                         stageName: 'First stage',
+                         jobName: 'Job 1',
+                         parameters: "[[\$class: 'StringParameterValue', name: 'foo', value: 'bar']]"
+                 ],
+                 [
+                         stageName: 'Second stage',
+                         jobName: 'Job 2',
+                 ]],
+        emails: ['jane@example.com', 'joe@example.com']
+).build(this);
+
+ppipelineJobStages.with {
+    logRotator {
+        numToKeep(365)
+    }
+}
+
+```
+
 ## JS Build Job
 
 ```groovy
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -92,46 +92,44 @@ new FlowJobBuilder(
 import jenkins.automation.builders.PipelineJobBuilder
 
 def script = """
-             node {
-                stage('First stage') {
-                    build job: 'Job 1',
-                    parameters: [
-                        [\$class: 'StringParameterValue', name: 'foo', value: 'bar'],
-                    ]
+    pipeline {
+        agent { label 'master' }
+        stages {
+            stage('hello') {
+                steps {
+                    sh 'echo "Hello World"'
                 }
             }
+        }
+    }
 """
 
-def pipelineJob = new PipelineJobBuilder(
-    name: 'Pipeline builder',
-    description: 'This is a simple pipeline job',
-    pipelineScript: script,
-    emails: ['jane@example.com', 'joe@example.com']
-
-).build(this);
-
-pipelineJob.with {
+new PipelineJobBuilder(
+        name: 'Hello Pipeline With Script',
+        description: 'This is a simple pipeline job',
+        pipelineScript: script
+).build(this).with {
     logRotator {
         numToKeep(365)
     }
 }
 
-def pipelineJobStages = new PipelineJobBuilder(
-        name: 'Pipeline builder',
+new PipelineJobBuilder(
+        name: 'Pipeline builder with stages',
         description: 'this is a simple pipeline job',
         stages: [[
-                         stageName: 'First stage',
-                         jobName: 'Job 1',
+                         stageName : 'First stage',
+                         jobName   : 'Job 1',
                          parameters: "[[\$class: 'StringParameterValue', name: 'foo', value: 'bar']]"
                  ],
                  [
                          stageName: 'Second stage',
-                         jobName: 'Job 2',
+                         jobName  : 'Job 2',
                  ]],
         emails: ['jane@example.com', 'joe@example.com']
-).build(this);
 
-ppipelineJobStages.with {
+).build(this)
+.with {
     logRotator {
         numToKeep(365)
     }
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -107,7 +107,8 @@ def script = """
 new PipelineJobBuilder(
         name: 'Hello Pipeline With Script',
         description: 'This is a simple pipeline job',
-        pipelineScript: script
+        pipelineScript: script, 
+        sandboxFlag: false
 ).build(this).with {
     logRotator {
         numToKeep(365)
@@ -125,9 +126,7 @@ new PipelineJobBuilder(
                  [
                          stageName: 'Second stage',
                          jobName  : 'Job 2',
-                 ]],
-        emails: ['jane@example.com', 'joe@example.com']
-
+                 ]]
 ).build(this)
 .with {
     logRotator {
--- a/example-jobs/pipeline_job.groovy
+++ b/example-jobs/pipeline_job.groovy
@@ -1,0 +1,45 @@
+import jenkins.automation.builders.PipelineJobBuilder
+
+def script = """
+
+node { stage 'Export latest task'
+build job: 'cf.gov-ansible-data-tasks',
+ parameters: [ [ \$class: 'StringParameterValue', name: 'EXPORT_ENV', value: EXPORT_ENV ],
+    [ \$class: 'StringParameterValue', name: 'IMPORT_ENV', value: IMPORT_ENV ],
+    [ \$class: 'StringParameterValue', name: 'TRANSFER_TASK', value: TRANSFER_TASK ] ]}
+
+"""
+def pipelineJob = new PipelineJobBuilder(
+        name: 'Pipeline builder',
+        description: 'this is a simple pipeline job',
+        pipelineScript: script,
+        emails: ['jane@example.com', 'joe@example.com']
+).build(this);
+
+pipelineJob.with {
+    logRotator {
+        numToKeep(365)
+    }
+}
+def pipelineJobStages = new PipelineJobBuilder(
+        name: 'Pipeline builder',
+        description: 'this is a simple pipeline job',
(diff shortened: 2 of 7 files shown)


example-jobs/pipeline_job.groovy
```diff
@@ -0,0 +1,48 @@
+import jenkins.automation.builders.PipelineJobBuilder
+
+def script = """
```
yunghuffy: This is cleaner when defined as stripping the top line, and not including the bottom line:

```
def script ="""\
...
}"""
```

example-jobs/pipeline_job.groovy
```diff
@@ -0,0 +1,48 @@
...
+    logRotator {
+        numToKeep(365)
+    }
+}
+
+def pipelineJobStages = new PipelineJobBuilder(
+        name: 'Pipeline builder',
+        description: 'this is a simple pipeline job',
+        stages: [[
+                         stageName: 'First stage',
```
acrewdson: wonky indentation here

src/main/groovy/jenkins/automation/builders/FlowJobBuilder.groovy
```diff
@@ -26,7 +26,9 @@ class FlowJobBuilder {
     List<String> emails
 
     /**
-     * @param DSL factory class, provided by Jenkins when executed from build context
+
+     * @param DSL factory class,  provided by Jenkins when executed from build context
```
acrewdson: somehow two spaces snuck in before 'provided'

src/main/groovy/jenkins/automation/rest/RestApiScriptRunner.groovy
```diff
@@ -3,7 +3,6 @@
 package jenkins.automation.rest
 
 import javaposse.jobdsl.dsl.DslScriptLoader
-import javaposse.jobdsl.dsl.ScriptRequest
```
acrewdson, line 6: not sure about this?

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * The basic PipelineJob builder
+ *
+ * @param stages list of maps describing the stages; see example linked below for the shape of the map
+ * @param name job name
+ * @param description job description
+ * @param pipelineScript optional string containing the raw pipeline script (use instead of `stages`)
+ * @param pollScmSchedule optional string in cron format to trigger builds on a scheduled interval
+ * @param emails list of email addresses to receive notifications
+ * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#TODOXXXXXXXXXXXX" target="_blank">Pipeline job builder example</a>
```
acrewdson: need to update the fragment identifier

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,87 @@
...
+                warningText += """
+                    ansiColor('xterm') {
+                        echo '\\033[31m! `pipelineScript` parameter has no effect because ' +
+                        '`stages` was also supplied; choose one or the other\\033[0m'
+                    }
+                """
+            }
+
+            it.description this.description
+            CommonUtils.addDefaults(delegate)
```
acrewdson: we have these two lines in here twice (see 34 and 35)

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,87 @@
...
+ * @param name job name
+ * @param description job description
+ * @param pipelineScript optional string containing the raw pipeline script (use instead of `stages`)
+ * @param pollScmSchedule optional string in cron format to trigger builds on a scheduled interval
+ * @param emails list of email addresses to receive notifications
+ * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#TODOXXXXXXXXXXXX" target="_blank">Pipeline job builder example</a>
+ */
+
+class PipelineJobBuilder {
+    List stages
```
acrewdson: I'd vote for `List<Map> stages` here

example-jobs/pipeline_job.groovy
yunghuffy: To elaborate, here's what I had in my test script

```
def script = """\
node {
  def buildSteps = [:]

  buildSteps['Build'] = {
    sh 'env | sort'
    echo 'Build step'
    sh 'uptime'
    echo '\$class'
  }

  buildSteps['Test'] = {
    sh 'echo teeeessssttttt'
    echo 'Test step'
    sh 'uptime'
  }

  parallel buildSteps
}"""
```

Which results in

image


If I have
```
def script = """
node {
  def buildSteps = [:]

  buildSteps['Build'] = {
    sh 'env | sort'
    echo 'Build step'
    sh 'uptime'
    echo '\$class'
  }

  buildSteps['Test'] = {
    sh 'echo teeeessssttttt'
    echo 'Test step'
    sh 'uptime'
  }

  parallel buildSteps
}
"""
```

Then I end up with
image

Notice the extra lines at the beginning and end.

imuchnik: Also notice escape slash before $class.

src/main/groovy/jenkins/automation/rest/RestApiScriptRunner.groovy
imuchnik: Unused import that got cleaned up by idea.

acrewdson: 👍

acrewdson (approved):
looks good to me 👍

example-jobs/pipeline_job.groovy
imuchnik: Yes, true, but all things considered, the functionality is the same. Slash may see a little more magical to non-groovy people, while stripIndent() is more expressive.  I am not inclined to add another commit just for that, if you don't object.

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,84 @@
...
+                    node {
+                        ${warningText}
+                        ${stagesStr}
+                    }
+                """.stripIndent()
+            }
+
+            definition {
+                cps {
+                    sandbox(false)
```
marcesher: Please make `sandbox` configurable

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,84 @@
...
+ * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#pipeline-builder" target="_blank">Pipeline job builder example</a>
+ */
+
+class PipelineJobBuilder {
+    List stages
+    String name
+    String description
+    String pipelineScript = ""
+    String pollScmSchedule
+    List<String> emails
```
marcesher: emails / extended email aren't actually an option on pipelines, so you'll need to remove this and the extended email block below

docs/examples.md
```diff
@@ -86,6 +86,59 @@ new FlowJobBuilder(
...
+                         parameters: "[[\$class: 'StringParameterValue', name: 'foo', value: 'bar']]"
+                 ],
+                 [
+                         stageName: 'Second stage',
+                         jobName: 'Job 2',
+                 ]],
+        emails: ['jane@example.com', 'joe@example.com']
+).build(this);
+
+ppipelineJobStages.with {
```
marcesher: typo (ppipeline). Also, I think we should encourage non-named jobs and just use the `.build(this).with {...}` convention

docs/examples.md
```diff
@@ -86,6 +86,59 @@ new FlowJobBuilder(
...
+
+).build(this);
+
+pipelineJob.with {
+    logRotator {
+        numToKeep(365)
+    }
+}
+
+def pipelineJobStages = new PipelineJobBuilder(
```
marcesher: If we're going to support this `stages` stuff then somewhere, it needs to be made clear why you'd choose this over just using pipelineScript. If I'm coming into this Builder documentation and I see two ways to do the same thing, with no explanation of the difference or why I'd choose one over the other, I'm just confused

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,77 @@
...
+ * @see <a href="https://github.com/cfpb/jenkins-automation/blob/gh-pages/docs/examples.md#pipeline-builder" target="_blank">Pipeline job builder example</a>
+ */
+
+class PipelineJobBuilder {
+    List stages
+    String name
+    String description
+    String pipelineScript = ""
+    String pollScmSchedule
+    Boolean sandboxFlag = true
```
yunghuffy, line 25: My one gripe/nitpick is that sandbox here is set to `true`. Looking through your past commits it looks like it was `false` by default. Any reason for this change? I don't think I've come across stuff we're using that expects this to be sandboxed.

acrewdson: It's a good point. I would have assumed `false` would be the default here. We don't want that?

acrewdson: I take it all back -- I think we're good here 😅  http://unethicalblogger.com/2017/08/03/donut-disable-groovy-sandbox.html

acrewdson:
I'd vote for merging 👍  -- and squashing when merging in this case

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,77 @@
...
+                triggers {
+                    scm pollScmSchedule
+                }
+            }
+
+            if (stages) {
+                def stagesStr = stages.collect {
+                    """
+                        stage("${it.stageName}") {
+                            build job: "${it.jobName}"
```
acrewdson, line 56: I think we need a comma at the end of this line -- see https://stackoverflow.com/a/39656390

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
acrewdson: just to be clear, sandboxing by default seems good to me

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -0,0 +1,77 @@
...
+                            ${it.parameters ? 'parameters: ' + it.parameters : ''}
```
acrewdson, line 57: I think we might need to call `.inspect()` on `it.parameters` to preserve the quoting

acrewdson (dismissed):
see comments

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
yunghuffy: Confirmed that parameters require the comma

```
[BFA] Scanning build for known causes...
[BFA] No failure causes found
[BFA] Done. 0s
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
WorkflowScript: 38: expecting '}', found ':' @ line 38, column 61.
   tranet-pulse-check' parameters: [boolean
                                 ^
```
from

```
    stages {
        stage('Build seed jobs'){
            steps {
                build job: 'parameterized job' parameters: [booleanParam(name: 'FOO_ME', value: true)]
            }
        }
    }
```

yunghuffy: Other thought, can we iterate over this when we actually need to use the `stages` part of this builder? I don't think we have been using this yet.

PR #109, opened 2017-10-04
Jira Card: DEV-2453-Fixing the sauce connect job bulder to use creden…

imuchnik:
…tils within job-dsl

also added options.

@marcesher @acrewdson

### DIFF
--- a/src/main/groovy/jenkins/automation/builders/SauceConnectJobBuilder.groovy
+++ b/src/main/groovy/jenkins/automation/builders/SauceConnectJobBuilder.groovy
@@ -11,7 +11,8 @@ import javaposse.jobdsl.dsl.Job
  * @param name job name
  * @param description job description
  * @param webDriverBrowser browser to use with Sauce Connect
- * @param sauceCredentialId jenkins id to use for the sauce plugin
+ * @param sauceCredentialId SauceCredential to use for the sauce plugin
+ * @param additionalOptions (Optional) additional option to use e.g --v
  *
  */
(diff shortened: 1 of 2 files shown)


src/main/groovy/jenkins/automation/builders/SauceConnectJobBuilder.groovy
```diff
@@ -50,16 +52,11 @@ class SauceConnectJobBuilder {
                     webDriverBrowsers(webDriverBrowser)
                     verboseLogging(true)
                     useGeneratedTunnelIdentifier(true)
+                    credentials(sauceCredentialId)
+                    additionalOptions? options (additionalOptions): null
```
acrewdson, line 56: with a ternary like this you usually will have spaces before and after the `?` and `:` operators, not just after; I think it'd be more readable that way

src/test/groovy/my_job_test.groovy
```diff
@@ -9,7 +9,7 @@ class my_job_test extends Specification {
         given:
         JobManagement jm = new MemoryJobManagement()
 
-        def testjob = new File('example-jobs/sqs_notifier_job.groovy').text
+        def testjob = new File('example-jobs/sauce_connect_job.groovy').text
```
acrewdson: it seems a bit strange that we keep changing this line in each PR to reflect the most recent thing being tested -- or am I missing something?

OrlandoSoto:
👍

PR #110, opened 2017-10-11
Fix so that parameters will now be passed (by adding required comma)

acrewdson:


src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -53,8 +53,9 @@ class PipelineJobBuilder {
                 def stagesStr = stages.collect {
                     """
                         stage("${it.stageName}") {
-                            build job: "${it.jobName}"
-                            ${it.parameters ? 'parameters: ' + it.parameters : ''}
+                            build job: "${it.jobName}"${
+                                it.parameters ? ', parameters: ' + it.parameters : ''
```
acrewdson, line 57: note that `inspect()` is not required here because params are being passed in as an already-escaped string (for better or worse)

imuchnik (approved):

imuchnik:
lgtm

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
yunghuffy: Does this preceding comma work?

acrewdson: @champain can you explain your question a bit more? are you asking if I tested this?

yunghuffy: Ok, gotcha offline. Makes more sense when I see the inputs.

yunghuffy (approved):

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
acrewdson: 👍

PR #113, opened 2017-10-11
DEV-2453 - Fix sauce connect job builder to use credentials

acrewdson:
(re-opening this since we reset and force pushed on master earlier, losing the commits [intentionally] 😬 )


OrlandoSoto (approved):
👍

PR #114, opened 2017-10-16
Copy `baseUrl` into the JENKINS_URL environment variable for the REST runner

marcesher:
We use the JENKINS_URL environment variable in several jobs.
When seed jobs run in the Jenkins runtime, Jenkins automatically
supplies this variable. However, when creating jobs using the REST
runner, the variable does not exist. This change ensures that
environment variable is available by setting it to the required
`baseUrl` parameter


OrlandoSoto (approved):
👍  thanks Marc!

acrewdson:
👍

PR #118, opened 2017-10-31
Untangles SauceConnectJobBuilder from JsJobBuilder

marcesher:
This removes the addition of a default NodeJS installation and removes a default archiveArtifacts.

Both of these changes are desirable, as they do not make sense as defaults on this builder

Addresses DEV-2478


imuchnik:
Looks good. Can we merge?

yunghuffy:
I can't find an example in my cursory look through job code where `node` is expressly called. I think this should be ok.

yunghuffy (approved):

PR #119, opened 2017-11-13
Checkmarx builder clean-up and refactoring

acrewdson:
This PR started as a few minor clean-up commits and ended with a larger refactoring commit that'll introduced breakages for existing Checkmarx JAC jobs.

Please step through the commits and review the diffs that way instead of reviewing the overall diff.


example-jobs/checkmarx_security_job.groovy
```diff
@@ -1,16 +1,15 @@
 import jenkins.automation.builders.CheckmarxSecurityJobBuilder
 
 def projectName ='foo'
-def groupId = "your-group-id"
+def groupId = "ac43cb0d-034d-4b1e-9bf5-e7c1c46f71d2"
```
acrewdson: NB: I am including this real-looking-guid-string on purpose so as to make it clearer what sort of value will be used for `groupId` in practice.

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
```diff
@@ -0,0 +1,172 @@
...
+        def excludeFoldersDefault = "resources, .git"
+        def incrementalDefault = true
+        def fullScanCycleDefault = "10"
+        def commentDefault = ""
+        def vulnerabilityThresholdDefault = true
+        def highThresholdDefault = 1
+        def mediumThresholdDefault = 2
+        def lowThresholdDefault = 3
+        def filterPatternDefault = """
+            !**/_cvs/**/*, !**/.svn/**/*,   !**/.hg/**/*,   !**/.git/**/*,  !**/.bzr/**/*, !**/bin/**/*,
```
yunghuffy: Because this is so large and hard to deal with have you considered moving it out of this file into a config file?

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
```diff
@@ -0,0 +1,172 @@
...
+
+                    if (cxConfig.useOwnServerCredentials) {
+                        'serverUrl'(cxConfig.serverUrl)
+                        'username'(cxConfig.username)
+                        'password'(cxConfig.password)
+                    }
+
+                    'projectName'(cxConfig.projectName)
+                    'groupId'(cxConfig.groupId) // really the Checkmarx 'team'
+                    'preset'(cxConfig.get("preset", presetDefault))
```
yunghuffy: Inconsistent quotes here. Seems like `'` was chosen, but then changed. Any reason?

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
```diff
@@ -0,0 +1,172 @@
...
+
+                    // TODO figure out whether this does anything
+                    'presetSpecified'(false)
+
+                    // TODO probably this should be configurable
```
champain: Seems like this functionality might be desirable. https://github.cfpb.gov/HMDA-Operations/hmda-devops/blob/master/jenkins/apps/env.groovy#L108 . Though they're not using this builder.

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
```diff
@@ -0,0 +1,172 @@
...
+
+                    'incremental'(cxConfig.get("incremental", incrementalDefault))
+
+                    // TODO is this right?
+                    'fullScansScheduled'(false)
+                    'fullScanCycle'(
+                        cxConfig.get("fullScanCycle", fullScanCycleDefault)
+                    )
+
+                    // TODO should we really be setting this?
```
yunghuffy: I thought group consensus was that incremental scans were faster. Don't we want to be able to do this?

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
```diff
@@ -0,0 +1,172 @@
...
+
+                    'projectName'(cxConfig.projectName)
+                    'groupId'(cxConfig.groupId) // really the Checkmarx 'team'
+                    'preset'(cxConfig.get("preset", presetDefault))
+
+                    // TODO figure out whether this does anything
+                    'presetSpecified'(false)
+
+                    // TODO probably this should be configurable
+                    'exclusionsSetting'('job')
```
yunghuffy: This seems more consistent with regards to quotes rather than the `element("value")` pattern I'm seeing elsewhere.

src/main/groovy/jenkins/automation/builders/CheckmarxSecurityJobBuilder.groovy
```diff
@@ -35,75 +25,22 @@ class CheckmarxSecurityJobBuilder {
...
-    String filterPattern = "!**/_cvs/**/*, !**/.svn/**/*"
-    String excludeFolders = "resources, .git"
-    String preset = "17"// Default 2014
-    Boolean presetSpecified = false
-    String fullScanCycle = "10"
-    Boolean vulnerabilityThresholdEnabled = true
-    String highThreshold = 1
-    String mediumThreshold = 2
-    String lowThreshold = 3
+    List scanRepo = []
```
yunghuffy, line 28: Does this parameter actually take multiple repos? What's the behavior here? All of our examples have just one repo.

yunghuffy (changes requested):
See file comments. I think there might be some easy style changes that _have_ to happen short of a good reason why not. The other comments are just seeking clarification.

imuchnik (approved):
👍

src/main/groovy/jenkins/automation/builders/CheckmarxSecurityJobBuilder.groovy
acrewdson: It does -- and this is not a change from before. See my PR in the operations-jenkins-automation repo for an example with multiple repos.

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
acrewdson: Not sure about a config file, but I can see them being moved out of the method, potentially. Is there an existing pattern here in JAC you can point to along these lines?

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
acrewdson: Just copy pasta from the old stuff. Will fix 👍

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
acrewdson: 😉

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
acrewdson: Definitely group consensus on that, and we are defaulting to incremental, see line 67. This is confusing: for some reason the Cx plugin creates this element and sets it to `false`, even when you have incremental enabled. I don't know why -- nor do I know what it does. Hence my comment. If you have any insights, lmk.

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
yunghuffy: I experimented with a couple of different methods, and I settled on the one I'm seeing in JAC from HMDA and from Delivery that uses a groovy config file then ConfigSlurper to parse it.

HMDA - https://[GHE]/HMDA-Operations/hmda-devops/blob/master/jenkins/apps/env.groovy
Delivery uses a config file in `jenkins-automation-private`

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
acrewdson: To be clear: this seems to have no impact on whether the scans are incremental, and we ARE defaulting to incremental, as best I can tell. But this was carried over from the older builder, and it is an element that's created (and set to `false`) when creating a Cx job via the Jenkins UI. No idea.

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
acrewdson: Cool, will take a look, thanks!

yunghuffy:
I just used this branch to deploy @laminfaty 's omnibus checkmarx job locally very successfully. I think these changes are sane and appear to be straightforward. Once the quotes are cleaned up, then it has my approval.

Great work.

acrewdson:
@jacob thanks for review and testing. I still plan to look into the ConfigSlurper stuff and give it a shot.

acrewdson:
@jacob good call on moving the defaults out of the method. I opted not to go for `ConfigSlurper`, wanting to keep the defaults pretty close and in the same file, so moved them to a map in the same class. I think this is an improvement. /cc @imuchnik @OrlandoSoto

yunghuffy:
My comment was about the large string of filter patterns being a little unwieldy. I'm good on this change, but modifying those filter patterns would be difficult in this code. That being said, it doesn't interfere with using this code to create jobs. So 👍

yunghuffy (approved):

acrewdson:
@champain ah, ok; FYI the large string of filter patterns is taken verbatim from the Checkmarx plugin -- it's the default value they provide. I chose not to try to reformat their string out of laziness.

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
```diff
@@ -0,0 +1,239 @@
...
+        excludeFolders: "resources, .git",
+        incremental: true,
+        fullScanCycle: "10",
+        comment: "",
+        vulnerabilityThresholdEnabled: true,
+        highThreshold: 1,
+        mediumThreshold: 2,
+        lowThreshold: 3,
+        osaEnabled: false,
+        osaIncludePattern: "**/osa_dependencies/**",
```
yunghuffy, line 219: Perhaps this needs to be documented?

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
```diff
@@ -0,0 +1,239 @@
...
+ *               <code>2</code>, sets the threshold for max number of 'medium'
+ *               vulnerabilities
+ * <br/>&#x2014; <code>lowThresholdDefault</code>: optional, defaults to
+ *               <code>3</code>, set the threshold for max number of 'low'
+ *               vulnerabilities
+ * <br/>&#x2014; <code>osaEnabled</code>: optional, defaults to
+ *               <code>false</code>, define whether or not to run
+ *               OSA dependency scans
+ * <br/>&#x2014; <code>osaIncludePattern</code>: optional, defaults to
+ *               <code>\*\*\/osa_dependencies\/\*\*</code>, set the pattern that OSA
```
yunghuffy, line 60: Wondering if this is significant enough documentation for the OSA scan path.

docs/examples.md
```diff
@@ -17,26 +17,76 @@ import jenkins.automation.builders.BaseJobBuilder
...
+            remote {
+                url('https://github.com/champain/jenkins-automation')
+            }
+            branch('checkmarx_osa_fixes')
+        }
+    }
+
+    steps {
+        gradle {
+            tasks('copyDependencies')
```
acrewdson, line 81: @champain this is pretty neat 👍

src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy
acrewdson: seems fine to me

OrlandoSoto (approved):
👍

yunghuffy:
Final 👍  on this one. I've been working with it for a week now without issue.

PR #120, opened 2017-11-14
Osa scaffold

yunghuffy:
This PR does three things

1. Adds optional OSA functionality to the current Checkmarx job.
2. Creates an example Checkmarx job with OSA.
3. Creates a utility in the build.gradle that can be used to download the dependency jars locally

Internally we were trying to decide if OSA would be part of normal checkmarx jobs or not, however OSA jobs rely on a lot of the internals of checkmarx jobs. They are indistinguishable, so reusing this code makes a lot of sense. These jobs also need to perform steps to download libraries before running OSA, so if this type of code isn't added in, then it might not be possible to simply extend this code in a job file.


imuchnik:
@champain I don't really follow all the dependencies. Can we huddle about it tomorrow?

yunghuffy:
@imuchnik Yes, I forget not everyone has been as involved with this part of Checkmarx, and we can definitely go over this.

yunghuffy:
There's a better way to do this.

PR #121, opened 2017-11-20
Allow gradle to generate wrapper

yunghuffy:
It's come to my attention that the gradle wrapper has been part of gradle's SOP for a while: https://docs.gradle.org/current/userguide/gradle_wrapper.html . It seems like these files are just generated by a task we're already including in the build.gradle.

To create the wrapper, all I had to do was `gradle wrapper`.


acrewdson:
@champain I'm confused here, since the docs you link to say `The Wrapper is something you should check into version control`, and they emphasize the word `should`. Is your interpretation of that something like, 'you should check in the ability to generate the wrapper, but not the wrapper script itself'?

yunghuffy:
So look at what's spurring this. We haven't updated gradle in years. The version we're dependent on is from 2014, and it's preventing us from running the version of groovy we'll actually be deploying this code to in Jenkins.

My real objection is the upgrade path. If I upgrade gradle and the wrapper, are you ok with me PRing all of that generated code? My instinct is that we don't 1. need to PR generated code and 2. don't want to PR generated code.

I'm certainly willing to listen to reason on this one, though.

acrewdson:
@champain I'm all for not checking-in the `gradlew` if we can avoid it (should we have some mention of how to generate it in the README?), but was just thrown off by that sentence and was curious what you made of it. Glad to follow your recommendation here -- you've been digging into it and I haven't. Also curious what @imuchnik thinks.

acrewdson (approved):
Seems good to me; would also like @imuchnik to take a look, though.

imuchnik:
It's been a while since I thought about it. I think we do need gradlew, but not gradle. We can upgrade the version of both groovy and gradle in 
build.gradle and gradle-wrapper.properties respectively.

UPDATE: I checked in the wrapper files in so that other dev would not need to install gradle. 
To generate the wrapper with gradle task you would ned to have gradle installed.  
Some docs here:
https://docs.gradle.org/current/userguide/gradle_wrapper.html

.gitignore
```diff
@@ -3,6 +3,10 @@
 build/
 sandbox/
 *.iml
+gradle/
+build/
+gradlew
```
imuchnik, line 8: We'd need both of those lines for developers who do not have gradle installed.

imuchnik (changes requested):

yunghuffy:
I think I see the usefulness of the gradle wrapper, and I understand that we can update libraries using it, but I think this is more of a philosophical thing. For instance if it installs gradle, then why not java, too?

I think I object generally to having a package manager in my git repos. I would rather have Homebrew, apt, yum, or pkg manage gradle then have a wrapper script create a consistent environment without worrying whether or not all of gradle's dependencies were installed and other package manager-like tasks.

I'm willing to accept group consensus on this, though.

yunghuffy:
I think there's little support for this movement. Closing.

PR #123, opened 2017-11-24
DEV-2502 - update default Checkmarx preset to use a newer one

imuchnik:
Changing default  preset for checkmarx


acrewdson (changes requested):
I think a bunch of other commits snuck into this PR. Also can you update the PR title?

PR #124, opened 2017-11-29
Move dependency downloader to importable init script

yunghuffy:
Adding this file allows a user to incorporate this function across gradle builds. Gradle can use this file when passing the `-I` flag.

### DIFF
--- a/gradle/init.d/init.gradle
+++ b/gradle/init.d/init.gradle
@@ -1,0 +1,18 @@
+def copyConfigs = {
+      def osa_dir = new File('osa_dependencies')
+      if ( !osa_dir.exists() ) {
+          osa_dir.mkdirs()
+      }
+      from project.configurations.compile
+      into 'jac_dependencies'
+}
+
+allprojects { project ->
+    afterEvaluate {
+        project.tasks.create(
+            name: 'copyDeps',
+            type: Copy,
+            copyConfigs,
+        )
+    }
+}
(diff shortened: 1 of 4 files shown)


gradle/init.d/init.gradle
```diff
@@ -0,0 +1,18 @@
+def copyConfigs = {
```
acrewdson, line 1: this looks good to me overall, only one question: what's the motivation for the name `copyConfigs` -- isn't this explicitly copying deps? or was there a name collision when creating the task below if you named the closure `copyDeps`?

yunghuffy: I had a real problem keeping the closure itself and the called task separate. When I was writing this sometimes it was hard to tell which context I was seeing errors in, the init script itself, or the task execution from gradle which has its own context.

acrewdson: ah, I see -- sounds like a pain; but now that you've got it working, what do you think about renaming it?

yunghuffy: I feel like the next person to work on this would hit the same issue I was hitting if they tried to work with this, especially if that person is me. What's the argument for renaming it?

acrewdson: no argument really -- if you think it's the right name, let's go with it 👍

gradle/init.d/init.gradle
```diff
@@ -0,0 +1,18 @@
...
+      def osa_dir = new File('osa_dependencies')
+      if ( !osa_dir.exists() ) {
+      osa_dir.mkdirs()
+          }
```
acrewdson: indentation is a bit weird in the two lines above

gradle/init.d/init.gradle
yunghuffy: Cool. If there is a legit namespacing issue or something to that effect, then I can totally be swayed. But if there's just a little merit to my debugging argument, then let's go with the little merit. Thanks for the feedback, though. Appreciate you keeping me honest.

acrewdson: sounds good 👍

yunghuffy:
This is also a merge when ready. Feedback from @imuchnik will be greatly appreciated.

acrewdson:
@champain good to merge?

yunghuffy:
This is ok to merge.

PR #127, opened 2018-01-03
Add configurable build result message and context

yunghuffy:
In order to complete this an additional feature was
  added that will break compatibility with any job
  using this library, however the use of a hash
  for parameters is too much of an improvement
  and this library is still lightly used.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -494,12 +494,15 @@ This PR watcher can be used to help configure jobs for either GitHub.com
 or a hosted GitHub Enterprise instance at a different hostname.
 
 Create a job that builds pull requests for a specific repo in a
-GitHub Enterprise instance. The `<unique github credential UUID>` should be
+GitHub Enterprise instance. The `<unique GH PR builder credential ID>` should be
 replaced with a credential ID created by the GitHub Pull Request Builder
 itself. This can be found in the `config.xml` of a job already configured
 to use the builder or from the `org.jenkinsci.plugins.ghprb.GhprbTrigger.xml`
 in a Jenkins installation.
 
+When running multiple PR Builder jobs on the same repo it's recommended to set
+a unique `ghPrStatusContext` so that all results appear as individual checks
+in the UI for the PR.
 
 #### For GitHub Enterprise
 
@@ -508,20 +511,30 @@ import jenkins.automation.builders.BaseJobBuilder
 import jenkins.automation.utils.GhUtils
 
 new BaseJobBuilder(
-    name: "GHE_PR_builder",
-    description: "Does GHE PR building",
+    name: 'GHE_PR_builder',
+    description: 'Does GHE PR building from internal GHE',
 ).build(this).with {
-    GhUtils.ghPrWatcher(delegate,
-                        'orgname/reponame',
-                        'github.organization.tld',
-                        '<unique github credential UUID>',
-                        true
+    GhUtils.ghPrWatcher(
+        delegate,
+        [
+            ghProject: 'cfpb/reponame',
+            ghHostname: 'github.org.tld',
+            ghAuthId: '<unique GH PR builder credential ID>',
+            ghPrHooks: true,
+            ghPrOrgsList: 'InternalDev',
+            ghPrStatusContext: 'Test your example job',
+            ghPrResultMessage: [
+              'SUCCESS': 'Tests completed normally',
+              'ERROR': 'Tests errored',
+              'FAILURE': 'Tests failed',
+            ]
+        ]
     )
 
     steps {
-        shell("""
-        echo 'Hello PRs!'
-        """.stripIndent()
+        shell('''
+        sh test.sh
+        '''.stripIndent()
         )
     }
 }
@@ -542,23 +555,31 @@ import jenkins.automation.builders.BaseJobBuilder
 import jenkins.automation.utils.GhUtils
 
 new BaseJobBuilder(
-    name: "GH_PR_builder",
-    description: "Builds PRs from public GitHub",
+    name: 'GH_PR_builder',
+    description: 'PR build from GitHub.com',
 ).build(this).with {
-    GhUtils.ghPrWatcher(delegate,
-                        'cfpb/example_repo',
-                        'github.com',
-                        '<unique github credential UUID>',
-                        false,
-                        false,
-                        '* * * * *',
-                        'cfpb'
+    GhUtils.ghPrWatcher(
+        delegate,
+        [
+            ghProject: 'cfpb/reponame',
+            ghHostname: 'github.com',
+            ghAuthId: '<unique GH PR builder credential ID>',
+            ghPrHooks: false,
+            ghPrCron: '*/2 * * * *',
+            ghPrOrgsList: 'CFPB',
+            ghPrStatusContext: 'Test your example job',
+            ghPrResultMessage: [
+              'SUCCESS': 'Tests completed normally',
+              'ERROR': 'Tests errored',
+              'FAILURE': 'Tests failed',
+            ]
+        ]
     )
 
     steps {
-        shell("""
-        echo 'Hello PRs from github.com'
-        """.stripIndent()
+        shell('''
+        sh test.sh
+        '''.stripIndent()
         )
     }
 }
--- a/example-jobs/ghe_job.groovy
+++ b/example-jobs/ghe_job.groovy
@@ -1,0 +1,32 @@
+import jenkins.automation.builders.BaseJobBuilder
+import jenkins.automation.utils.GhUtils
+
+new BaseJobBuilder(
+    name: 'GH_PR_builder',
+    description: 'Does GH PR building',
+).build(this).with {
+    GhUtils.ghPrWatcher(
+        delegate,
+        [
+            ghProject: 'cfpb/reponame',
+            ghHostname: 'github.org.tld',
+            ghAuthId: '<unique GH PR builder credential ID>',
+            ghPrHooks: false,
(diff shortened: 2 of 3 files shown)


src/main/groovy/jenkins/automation/utils/GhUtils.groovy
```diff
@@ -12,51 +14,87 @@ class GhUtils {
...
-                    if (ghPrCron) {
-                        spec(ghPrCron)
-                        cron(ghPrCron)
-                    }
-                    else {
-                        spec()
-                        cron()
-                    }
+                    // Avoid 'null' in these blocks
+                    spec(ghPrConfig['ghPrCron'] ?: '')
```
yunghuffy: This is a bit baffling, but if I try to look up the value via `get` for this particular node, then the rest run bombs out with the following:

```
No signature of method: groovy.util.NodeBuilder.spec() is applicable for argument types: (java.lang.String, java.lang.String) values: [*/2 * * * *, ]
```
I'm not 100% sure why that's happening since it works with this style look up. The Elvis operator is in there because it's cosmetically bad to leave 'null' in the config, it creates unnecessary diffs, I'm not 100% sure how XML handles the string 'null', and it's not valid cron spec.

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
```diff
@@ -12,51 +14,87 @@ class GhUtils {
...
+                            triggeredStatus()
+                            startedStatus()
+                            statusUrl()
+                            addTestResults(true)
+                            def ghPrMessage = ghPrConfig.get(
+                                'ghPrResultMessage',
+                                defaults.ghPrResultMessage
+                            )
+                            def confCompletedStatus = completedStatus()
+                            ghPrMessage.each{ buildStatus, ghMessage ->
```
yunghuffy, line 81: I considered moving this to its own function, but I couldn't really see this getting reused.

yunghuffy: I also debated not putting these options in a map since there's just three of them, this just _looked_ cleaner to me. Open to suggestions.

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
```diff
@@ -12,51 +14,87 @@ class GhUtils {
...
- * @param  ghHostname The GitHub hostname, i.e. github.org.tld, or github.com
- * @param  ghAuthId the Jenkins credential ID for GitHub authentication
- *         see <a href="https://github.com/jenkinsci/ghprb-plugin/blob/master/README.md#credentials">the plugin docs</a>
- * @param  ghPermitAll  optional Whether or not to allow all contributors to build a PR
- * @param  ghPrHooks  optional Whether or not to install webhooks in the remote repo
- * @param  ghPrCron  optional For use when ghPrHooks is false. Uses standard cron syntax
- * @param  ghPrOrgList  optional A string of organizations to whitelist for PRs
+ * @param  ghPrConfig A <code>Map</code> of configuration for the PR builder.
+                      Valid keys:
+ * <br/>&#x2014;  <code>ghProject</code> The GitHubg project name, e.g. username/reponame
```
JeffreyMFarley: Typo

JeffreyMFarley (approved):
I can't wait

docs/examples.md
```diff
@@ -508,20 +511,30 @@ import jenkins.automation.builders.BaseJobBuilder
...
-                        'orgname/reponame',
-                        'github.organization.tld',
-                        '<unique github credential UUID>',
-                        true
+    GhUtils.ghPrWatcher(
+        delegate,
+        [
+            ghProject: 'cfpb/reponame',
+            ghHostname: 'github.org.tld',
+            ghAuthId: '<GitHub user UUID>',
```
acrewdson: I'm finding it's pretty hard to describe what this ID is, but in an accurate and concise way. Isn't this more the ID of the auth configured for the ghprb plugin? -- that bit under `<githubAuth>` in `org.jenkinsci.plugins.ghprb.GhprbTrigger.xml`? But that's pretty hard to say in the space allotted here. Any ideas for coming up with something a little more descriptive?

yunghuffy: It's the 'GitHub Pull Request Builder credential ID.' Seems wordy, but perhaps clearer.

docs/examples.md
```diff
@@ -542,23 +555,31 @@ import jenkins.automation.builders.BaseJobBuilder
 import jenkins.automation.utils.GhUtils
 
 new BaseJobBuilder(
-    name: "GH_PR_builder",
-    description: "Builds PRs from public GitHub",
+    name: 'GHE_PR_builder',
```
acrewdson: should this just be `GH_PR_builder`?

example-jobs/ghe_job.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+new BaseJobBuilder(
+    name: 'GHE_PR_builder',
+    description: 'Does GHE PR building',
+).build(this).with {
+    GhUtils.ghPrWatcher(
+        delegate,
+        [
+            ghProject: 'cfpb/reponame',
+            ghHostname: 'github.com',
```
acrewdson: if this is GHE, should this be a different hostname?

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
acrewdson: what about `ghPrConfig.get('ghPrCron', '')` instead?

acrewdson (approved):
great job -- looks good!

example-jobs/ghe_job.groovy
```diff
@@ -0,0 +1,32 @@
...
+              'SUCCESS': 'Tests completed normally',
+              'ERROR': 'Tests errored',
+              'FAILURE': 'Tests failed',
+            ]
+        ]
+    )
+
+    steps {
+        shell('''
+        sh test.sh
```
imuchnik: I think this will fail our travis build, why not just use 'echo'?

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
```diff
@@ -12,51 +14,87 @@ class GhUtils {
...
- * @param  ghHostname The GitHub hostname, i.e. github.org.tld, or github.com
- * @param  ghAuthId the Jenkins credential ID for GitHub authentication
- *         see <a href="https://github.com/jenkinsci/ghprb-plugin/blob/master/README.md#credentials">the plugin docs</a>
- * @param  ghPermitAll  optional Whether or not to allow all contributors to build a PR
- * @param  ghPrHooks  optional Whether or not to install webhooks in the remote repo
- * @param  ghPrCron  optional For use when ghPrHooks is false. Uses standard cron syntax
- * @param  ghPrOrgList  optional A string of organizations to whitelist for PRs
+ * @param  ghPrConfig A <code>Map</code> of configuration for the PR builder.
+                      Valid keys:
+ * <br/>&#x2014;  <code>ghProject</code> The GitHub project name, e.g. username/reponame
```
imuchnik, line 19: What are all these extra chars and markup. I prefer that these groovydoc comments be also readable in code.

imuchnik (changes requested):

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
yunghuffy: I tried that and preferred it, but it was throwing the above error. Let me know if you get that working in testing.

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
yunghuffy: These are bullet points when rendered in HTML, and they're consistent with https://github.com/cfpb/jenkins-automation/blob/master/src/main/groovy/jenkins/automation/utils/CheckmarxUtils.groovy where I borrowed them. @imuchnik  are you objecting to all markup or just these bullets?

example-jobs/ghe_job.groovy
yunghuffy: I don't see the Travis build running anything other than:

```
  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
```
And the build appears to have passed.

It's not even performing any gradle steps in this repository. Should it be running `build` or something? I definitely don't understand the idea of rendering and also running Jenkins jobs via Travis in this repo's current state.

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
acrewdson: @champain I just tried it locally, no problems. The error you included above makes it look like there was something else going on, possibly. Can you try again and see if it works?

src/main/groovy/jenkins/automation/utils/GhUtils.groovy
imuchnik: Just the bullets. I think 
```
@param paramName paramDescription
```
should be sufficient.

docs/examples.md
```diff
@@ -542,23 +555,31 @@ import jenkins.automation.builders.BaseJobBuilder
...
+            ]
+        ]
     )
 
     steps {
-        shell("""
-        echo 'Hello PRs from github.com'
-        """.stripIndent()
+        shell('''
+        sh test.sh
```
imuchnik, line 581: I think this will break our Travis build, since test.sh does not exist.

PR #130, opened 2018-01-24
DEV-2567 - small improvements to MultibranchPipelineJobBuilder

acrewdson:


src/main/groovy/jenkins/automation/builders/MultibranchPipelineJobBuilder.groovy
```diff
@@ -1,55 +1,69 @@
...
+                            remote(gitRemote)
+                            credentialsId(gitCredentials)
+                        }
+                        break
+                    case BranchSourceType.GITHUB:
+                        github {
+                            scanCredentialsId(gitCredentials)
+                            repoOwner(ghOwner)
+                            repository(ghRepo)
+                            apiUri(ghApiEndpoint)  // TODO this does not seem to work
```
acrewdson: ☝️

src/main/groovy/jenkins/automation/builders/MultibranchPipelineJobBuilder.groovy
```diff
@@ -1,55 +1,85 @@
...
-    String gitRepository
-    String gitEndpoint
+    String ghOwner
+    String ghRepo
+    String ghApiEndpoint
     String gitRemote
     int oldNumToKeep = 10
 
+    enum BranchSourceType {
+        /**
```
yunghuffy, line 47: Style choice on multi-line here?

src/main/groovy/jenkins/automation/builders/MultibranchPipelineJobBuilder.groovy
```diff
@@ -1,55 +1,85 @@
...
-import jenkins.automation.utils.Branch
-
 import javaposse.jobdsl.dsl.DslFactory
 import javaposse.jobdsl.dsl.jobs.MultibranchWorkflowJob
-import jenkins.automation.utils.Branch
+
+
+/*
+    NB: if you are polling via the GitHub API (github.com or GHE) and
+        want to use webhooks with a multi-branch pipeline, but the webhooks
```
yunghuffy: 'multi-branch' or "multibranch?' Either way it will help to be consistent.

src/main/groovy/jenkins/automation/builders/MultibranchPipelineJobBuilder.groovy
acrewdson: good question; doing it that way was the only way I could get these comments to be included as `description`s in the groovydoc

yunghuffy: This was not a pun, I swear.

yunghuffy (commented):
Overall I'm concerned that we're not leading the horse to the gentle water that is setting up the GHE/GH servers when necessary in the `/configure` section of Jenkins before rolling this out.

src/main/groovy/jenkins/automation/builders/MultibranchPipelineJobBuilder.groovy
acrewdson: good call; no hyphen I believe is the correct one, will fix

acrewdson:
@champain not sure I follow the horse/water stuff -- can you explain more what changes you're looking for? I tried to make some of that explicit in the groovydoc string on the class, but I think there's a limit to how much we can put about specific Jenkins config in this context

OrlandoSoto (approved):

yunghuffy:
I thought the idea was to tell people that they also likely need to configure a GHE instance and a GH instance in the Jenkins `configure` screen to get this to work on GHE in order to generate web hooks and work out of the box. We saw that you can use this builder, point it at a GHE instance and get a job to build, but not be in a good state if you don't literally configure the GHE and GH instance you're trying to talk to, and I don't think that's clear from the document strings.

yunghuffy (approved):

PR #132, opened 2018-02-15
DEV-2600 Update Sauce Connect builder and add example

OrlandoSoto:
We recently discovered that `sauceCredentialId` is not populated when the job is created using Gradle build. However, we do not see this issue when the job is created via the seed job inside Jenkins.
It seems that when we use Gradle build the code does not fall inside this conditional block: https://github.com/jenkinsci/job-dsl-plugin/blob/master/job-dsl-core/src/main/groovy/javaposse/jobdsl/dsl/helpers/wrapper/WrapperContext.groovy#L650-L652

The job-dsl uses `jobManagement.isMinimumPluginVersionInstalled` to determine if the version of the Sauce On Demand Jenkins plugin is `1.148` or higher. This does not seem to be available via Gradle build.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -259,6 +259,22 @@ new SalesforceAntJobBuilder(
 
 ```
 
+## Sauce On Demand job
+```groovy
+import jenkins.automation.builders.SauceConnectJobBuilder
+
+def projectName ='foo'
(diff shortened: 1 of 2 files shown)


acrewdson (approved):
👍 thanks @OrlandoSoto

PR #135, opened 2018-07-18
Upgrade gradlew and groovy

imuchnik:
Resolves #706


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.8-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
```
marcesher, line 6: @imuchnik is `4.1` the correct version number to use? I ask because that version is almost a year old, and it seems we should update to the latest if possible. What do you think?

build.gradle
```diff
@@ -13,8 +14,7 @@ apply plugin: 'maven'
 repositories {
     mavenCentral()
     jcenter()
-    maven { url 'http://repo.jenkins-ci.org/releases/' }
-}
+    maven { url 'http://repo.jenkins-ci.org/releases/' }}
```
acrewdson: Could the second `}` go back on its own line, like before?

acrewdson (changes requested):
There are some odd commits in this PR. Could you fix the PR to remove those?

gradle/wrapper/gradle-wrapper.properties
imuchnik: We can potentially go  to 4.2, but starting at 4.2.1 there are breaking changes that will not work without some more time spent on code.

marcesher:
Thanks for explaining the version, @imuchnik . I've pulled this locally and all works well.

Please merge when you all are ready and able to monitor for effects this will have on all our seed jobs.

PR #136, opened 2018-09-10
Remove FlowJobBuilder and associated examples

acrewdson:
The Jenkins Build Flow plugin is no longer maintained and has several known security vulnerabilities. With the implementation of JEP-200 earlier this year, it's now even more necessary to abandon support for Build Flow jobs, since the plugin is not being updated for JEP-200 compatibility. For these reasons, it makes sense for us to now remove the `FlowJobBuilder` class and related bits of example code. It has been recommended for some time to use Jenkins Pipelines in place of Build Flow jobs.

See:
* [Build Flow plugin](https://wiki.jenkins.io/display/JENKINS/Build+Flow+Plugin)
* [JEP-200](https://jenkins.io/blog/2018/01/13/jep-200/)
* [Plugins affected by JEP-200](https://wiki.jenkins.io/display/JENKINS/Plugins+affected+by+fix+for+JEP-200)


OrlandoSoto (approved):
👍 thanks Andrew!

PR #140, opened 2019-05-13
removing addDefaults from the PipelineJobBuilder

Josh-thephillipsequation:


JimRamia:
This looks good.  Don't see a merge option though for me.  Someone else will ahve to merge.

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
```diff
@@ -32,7 +32,9 @@ class PipelineJobBuilder {
     Job build(DslFactory factory) {
         factory.pipelineJob(name) {
             it.description this.description
-            CommonUtils.addDefaults(delegate)
+            logRotator {
```
imuchnik, line 35: Why was this line changed?

Josh-thephillipsequation: This resolves issue #140 (basically parts of the addDefaults method were deprecated in the new version of the job DSL plugin

imuchnik: https://github.com/cfpb/jenkins-automation/blob/master/src/main/groovy/jenkins/automation/utils/CommonUtils.groovy#L14  We should fix the method. This is mostly for uniformity, so we can change that logRotator value if we need to.

imuchnik (changes requested):
We need to fix it in utils, so that it works everywhere it is used

imuchnik:
I think this PR needs to be closed.

src/main/groovy/jenkins/automation/builders/PipelineJobBuilder.groovy
Josh-thephillipsequation: this was the change requested by @marcesher as the common utils method didn't have any additional non deprecated features beyond the logRotator

imuchnik (approved):

PR #142, opened 2019-05-29
Update job-dsl dependency to latest

marcesher:
We're now running the latest job-dsl-plugin on our Jenkins servers. This
gets our automation helpers up to that version


imuchnik:
Let's figure out why it is failing

marcesher:
replacing with a new PR

PR #143, opened 2019-10-04
APPOPS-425 Add id parameter to multibranch pipeline builder

OrlandoSoto:
- Add `id` parameter to branchSources. This is required parameter in job-dsl plugin 1.76

The `id parameter` is set to correspond to the job name, in order to keep this parameter constant and unique within the Jenkins instance


marcesher:
@OrlandoSoto this makes sense to me as a starting point. We may learn sometime down the road that we might need to make it a parameter, but I think it makes sense to not do that now and go with the default of `name`. Thanks!

PR #144, opened 2019-10-07
Disable concurrency util

marcesher:
See commit comments and comment on new CommonUtils.disableConcurrentBuilds() method

@imuchnik and I discussed this prior to the PR and she knows why we're doing this and saw the code I intended to write


marcesher:
Because I ran this by @imuchnik first, I'm going ahead and merging this

PR #146, opened 2019-10-22
Add credentials binding util

marcesher:
Adds a utility to configure usernamePassword and amazonWebServices credentials without needing job-dsl dynamic methods

The problem with job-dsl's dynamic methods is that they do not work in the context of `gradlew test` and `gradlew rest`. This addition enables us to use clean usernamePassword and amazonWebServices credentials in our jobs and still retain gradle-based unit and rest test workflows.

Improves APPOPS-444 considerably.


OrlandoSoto:
LGTM I was able to build/run locally by adding `some-credentials-id` and `some-aws-credentials-id` on my local Jenkins credentials store

OrlandoSoto (approved):
👍  thanks Marc!

